PR #9, opened 2015-09-13
fix validation error, file specification not associated with an object

inyenia:
PDFA3FileAttachment.pdf does not conform to PDF/A.

Validate PDF/A 3-a File online at http://www.pdf-tools.com/pdf/validate-pdfa-online.aspx

I found how to solve the problem at https://www.mail-archive.com/search?l=users@pdfbox.apache.org&q=subject:%22How+can+I+embed+files+in+PDF%2FA-3a%3F%22&o=newest&f=1

I hope you find it useful


jstaerk:
My PDF/A-3 (ZUGFeRD) file attachments seem to work with the mentioned validator?

PR #13, opened 2017-02-02
use PDFbox2, log4j2, gradle

dovogt:
Change from pdfbox1 ->pdfbox2
Change from jempbox ->xmpbox
use Log4j2 as Logger
check sources with Checkstyle, FindBugs and PMD
change to java conform source files
build with gradle


dovogt:
log4j2 config

Issue #17, opened 2017-05-09
Removing temporal couplings from ZUGFeRDExporter to make it more intuitive


yankee42:
The class `ZUGFeRDExporter` has a number of methods which only have a meaning at a certain point in time. For example:
```java
ZUGFeRDExporter zugFeRDExporter = new ZUGFeRDExporter();
zugFeRDExporter.PDFmakeA3compliant(/* .. */);
zugFeRDExporter.setConformanceLevel("A"); // Meaningless - import already happened
zugFeRDExporter.ignoreA1Errors(); // Meaningless - import already happened
```
Calling some methods early will cause errors:
```java
ZUGFeRDExporter zugFeRDExporter = new ZUGFeRDExporter();
zugFeRDExporter.export("myFile"); // NullPointerException
zugFeRDExporter.PDFattachZugferdFile(/* ... */); // NullPointerException
```
In my opinion this makes the API less intuitive. Also it makes this class quite huge, because it is doing several unrelated tasks (such as loading and what comes after that).

I'd propose to fully initialize the document in the constructor and perform any loading tasks in factory methods or a separate factory class. I created a draft for that [her](https://github.com/yankee42/mustangproject/tree/exporter-factory). This would change the API as follows:

1. Simply loading a file:
    ```java
    // OLD:
    ZUGFeRDExporter zugFeRDExporter = new ZUGFeRDExporter();
    zugFeRDExporter.loadPDFA3("myFile.pdf");
    
    // NEW:
    ZUGFeRDExporter zugFeRDExporter =
        ZUGFeRDExporter.createFromPDFA3("myFile.pdf");
    ```
2. Load a PDF/A1 file with default options:
    ```java
    // OLD:
    ZUGFeRDExporter zugFeRDExporter = new ZUGFeRDExporter();
    zugFeRDExporter.PDFmakeA3compliant("myFile.pdf", "a", "b", true)
    
    // NEW:
    ZUGFeRDExporter zugFeRDExporter =
        new ZUGFeRDExporterFactory().createFromPDFA1("myFile.pdf", "a", "b", true);
    ```
3. Loading a PDF/A1 file with extra options
    ```java
    // OLD:
    ZUGFeRDExporter zugFeRDExporter = new ZUGFeRDExporter();
    zugFeRDExporter.setConformanceLevel("A");
    zugFeRDExporter.ignoreA1Errors();
[...]

PR #18, opened 2017-05-10
Proposal on changes of the ZUGFeRDExporter-API

yankee42:
In extension to #17 I'd propose to change the API of the `ZUGFeRDExporter`. The changes in this pull request should be backward compatible, so that users have time to adjust.

The changes are as follows:

1. I extracted the `ZUGFeRDExporterFromA1Factory` from the `ZUGFeRDExporter`. I already outlined the reasons for that in #17. In short: This removes temporal couplings, which force the user to call methods in a specified order without the order being fully apparent: Settings which only play a role for PDF conversion are only available when actually performing a conversion and also only before doing the conversion.
2. I removed JAXB exceptions from public API to improve encapsulation. I think the user of the library should not know which method is used to serialize the objects to XML. This also makes it possible to exchange/upgrade the XML generation library without changing the public API in case this becomes necessary. As far as I can tell there is nothing a user could do about a JAXBException if he caught one, because he has no influence of the usage.
3. I used enums as configuration constants, so that it is always obvious which choices there are and to check these during compile time.
4. I improved exception handling at a few locations: I don't think that IOExceptions should be caught by mustang. If an Exception is caught like "hard disk full" or "could not read file", then this is in the domain of the user of the library to solve or not to solve. Especially this is a problem if an Exception is caught, possibly logged somewhere where the user won't see it and follow up errors occur. E.g. if the `isValidA1` method cannot read the the input PDF file (because the file was deleted in the meantime, the network storage broke down,...) then it supresses the error and returns instead that the document is not valid A1 (although the method could not check that). Since the error was suppressed the application continues and follow-up errors might arise that's origin is hard to determine. Also in the method `getA1ParserValidationResult` I moved the closing of the document into a finally block, to ensure that it is always closed, even in the case of an exception.

Please let me know what you think.

### DIFF
--- a/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDConformanceLevel.java
+++ b/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDConformanceLevel.java
@@ -1,0 +1,5 @@
+package org.mustangproject.ZUGFeRD;
+
+public enum ZUGFeRDConformanceLevel {
+    BASIC, COMFORT, EXTENDED
+}
--- a/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExportException.java
+++ b/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExportException.java
@@ -1,0 +1,18 @@
+package org.mustangproject.ZUGFeRD;
+
+public class ZUGFeRDExportException extends RuntimeException {
+    public ZUGFeRDExportException() {
+    }
+
+    public ZUGFeRDExportException(String message) {
+        super(message);
+    }
+
+    public ZUGFeRDExportException(String message, Throwable cause) {
+        super(message, cause);
+    }
+
+    public ZUGFeRDExportException(Throwable cause) {
+        super(cause);
+    }
+}
--- a/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExporter.java
+++ b/mustang/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExporter.java
@@ -26,10 +26,7 @@ import java.util.GregorianCalendar;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.logging.Level;
-import java.util.logging.Logger;
 
-import javax.activation.FileDataSource;
 import javax.xml.bind.JAXBContext;
 import javax.xml.bind.JAXBElement;
 import javax.xml.bind.JAXBException;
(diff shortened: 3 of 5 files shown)


jstaerk:
I'm in the process of releasing mustang 1.4.0, I will review the changes afterwards

yankee42:
OK. I just rebased on current master. Let me know when you are done with the new release, so I'll rebase again.

BTW, I just noticed, that in commit 9d2e4f205bdfdc79b4cdac166afc8c8b359e52b8 you set a default behaviour in case the conformance level is null to EXTENDED. However in commit 8389395451d7fa6a7de56bcd07a63d47b9a2160d this was removed (intentional or by accident?). I "added" that behaviour again, by setting default values in the new factory that I created.

jstaerk:
1) 1.4.0 is now released
2) I like in particular your numbers 2, 3 and 4
3) I think the re-introduction of "extended" was accompanied by again using that member but I'm not sure and if it is not used it should indeed be removed.
4) AFAIK PDFAConformanceLevel "U" is short for Unicode, I think the order is A (fully accessible), U (text accessible) B (not accessible).
5) please reintroduce isValidA1 as explained in #17 .
6) fundamentally, several small merge requests would be faster to process because I rather have 4 quarter of an hour slots than one one hour slot for review :-)

jstaerk:
PS: If you are interested in ZUGFeRD and Open Source feel free to join the next ZOSOCON on May 26 https://github.com/ZUGFeRD/ZOSOCON

yankee42:
Great to here. I'll be glad to split the whole thing in smaller pull requests. So I'll create one for introduction of enums, one for removing JAXB-Exceptions and one for the Exceptionhandling. I'll wait for your input on #17 before doing the same for the factory.

jstaerk:
ok, sorry, somewhere it looked as if the pdf validation had been removed altogether but I probably missed a insert at another place. I merged the sub-PR. Apologies if you don't hear anything for some days, I will be preparing and attending zugferd-entwicklertage.de but maybe I find a quiet hour not only to update the documentation and the maven but also to clear the conflict and test your branch.

yankee42:
Too bad, that I cannot be there :-(. But don't worry, I think I have never in my life created pull requests that were processed that quickly ;-). I will be happy to clear the conflicts myself, as soon as we are good with the sub-PRs.

jstaerk:
please resolve the conflicts. And it seems as if you have deleted the versionStr. It's a hassle to maintain and it would be great if the value could be taken from the POM but I would not like having it removed altogether.

yankee42:
Thanks for merging. However, maybe that is a misunderstanding: I did not resolve the conflicts yet and now the project does not build anymore due to the conflicts. I wanted to wait with resolving the conflicts until we are done with the #19, #20 and #22, because otherwise that would require more work, if I don't base it on the final of those.

What would you like to do and how can I help? I'd propose to undo the merge (that is: reset master on 58d930b), then I'd resolve the conflicts and you can merge again? Or do you prefer something else?

PR #26, opened 2017-05-29
Use version from pom.xml instead java constant

yankee42:
You [previously mentioned](https://github.com/ZUGFeRD/mustangproject/pull/18#issuecomment-303210373) that you'd prefer to read the version number from the maven POM file rather than maintaining an additional constant in the project.

I think that makes sense. This pull request will do just that. When compiling with maven, it will add a class `Version` with a constant `VERSION` to the target/generated-sources directory, which can then be accessed normally.

Note that your IDE might complain about the missing file (although the project will build normally with maven): Build the project once with maven, so that maven generates the version class and then, in your IDE, mark the directory `target/generated-sources/java-templates` as "generated sources root" (IntelliJ IDEA) or only "src" or whatever you IDE uses as designator.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -120,6 +120,19 @@
           </execution>
         </executions>
       </plugin>
+      <plugin>
+        <groupId>org.codehaus.mojo</groupId>
+        <artifactId>templating-maven-plugin</artifactId>
+        <version>1.0.0</version>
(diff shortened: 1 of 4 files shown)


jstaerk:
cool thanks

PR #27, opened 2017-05-29
Cleanup ZUGFeRDExporter after merge errors

yankee42:
As a followup of #18: There were a couple of merge conflicts and the resolution of those were not quite complete. This PR fixes what went wrong during the merge.

This is in summary:
- The enum `PDFAConformanceLevel` was contained twice in the codebase in different packages. This PR deletes the one in package `model`.
- A lot of code was originally moved from the class `ZUGFeRDExporter` to the new class `ZUGFeRDExporterFromA1Factory`. While the new class survived the merge, the code is now duplicated in both the new class and the old `ZUGFeRDExporter`. This class removes the moved code from `ZUGFeRDExporter` in a backward compatible way and marks the old code as deprecated
- The unit tests are now also adapted, so that they do not use the deprecated methods anymore.
- As a follow-up to #16 the unit tests now also close the exporter.


yankee42:
I just rebased this PR on master (and resolved a conflict with PR #26 in the process)

PR #33, opened 2017-06-02
close resource leaks and use tryWithResource

dovogt:


yankee42:
This PR caused the build to break, due to compilation errors, because in the pom.xml java 6 is specified as language level. But even that aside: The ressources closed were (at least partially) no leaks. The `ZUGFeRDExporterFromA1Factory` will now give a closed document to the `ZUGFeRDExporter` which will then be inable to do the exporting. So `ZUGFeRDExporterFromA1Factory` is now unable to function.

So what to do? Should the pom.xml be upgraded to Java 7 or Java 8? Revert the PR and fix it first? Or fix it now?

dovogt:
The resource leak is noticed when importing invoices. Because of this, I have looked at all the sites again and revised. It is not a nice procedure to not close resources.

jstaerk:
language level should be 1.7

PR #56, opened 2018-05-01
Improving Version Migration+Tests & Copyright Header to all files

svanteschubert:
Two improvements one in each commit:

1)
Regarding the migration ZUGFeRD1->2 feature: 
Allowing transformation of complete input directory, with reference check for any ref file existent with line base comparison for better usability. Including several refactoring and tweaks.

The JavaDoc of the VersionMigrationTest writes:
/**
 * Every ZUGFeRD 1.0 XML file from <code>src/test/resources/migration/input</code>
 * will be migrated to <code>src/test/resources/migration/output</code>.
 * If there is a similar named file as test reference <code>src/test/resources/migration/reference</code> a test
 * will be triggered for every single file.
 *
 * Note: Any 'ZUGFeRD1' will be exchanged to 'ZUGFeRD2' during migration and adequate reference will be searched for.
 */

2) 
Adding AL2 copyright header to Java and XML files with Jochen's copyright.


jstaerk:
optically looks ok, thanks for the migrate directory feature. I merged it but I will test it much later

PR #78, opened 2019-04-02
Add BLZ&Kto.-Nr

HardyLoppmann:


jstaerk:
hi @HardyLoppmann 
We have to talk: pg 31 of the spec says ZF2 no longer supports Bankleitzahl as payment mean, and apparently it violates the schema file. We can leave it for ZF1 but unless you tell me how to make it compliant to the schema file I will remove it for ZF2.

HardyLoppmann:
I think you should remove it for ZF2

Issue #90, opened 2019-06-18
PDF/A-3 conversion missing in ZUGFeRDExporter.export(OutputStream output)


tarikosman:
If the ZUGFeRD attachment is added using `PDFAttachGenericFile(...)` and exported to an OutputStream, the result is not PDF/A-3 compliant. Apparently only `org.mustangproject.ZUGFeRD.ZUGFeRDExporter(String ZUGFeRDfilename)` contains a call to `prepareDocument();` which is missing in export(OutputStream output).

jstaerk:
Since it's possible that the user already has a PDF/A-3, the according conversion should only take place if invoked via ZUGFeRDExporterFromA1Factory(). Anyway, if you already identified the cause, could you send a PR with the patch and ideally a test case?

swsch:
I _do_ have a PDF/A-3b file as input and the following code _does not_ produce a PDF/A-3u file:
```
ZUGFeRDExporter ze = new ZUGFeRDExporterFromA3Factory()
                    .ignorePDFAErrors()
                    .setZUGFeRDVersion(2)
                    .setZUGFeRDConformanceLevel(ZUGFeRDConformanceLevel.EN16931)
                    .load("2019-1265.pdf");
            // ze.setPDFA3(true); // 1
            ze.PDFattachZugferdFile(this);
            // ze.setPDFA3(true); // 2
            ze.export("2019-1265_z.pdf");
```
However, after changing the `A3Factory` to an `A1Factory` the output file conforms to PDF/A-3u.

Uncommenting the line `// ze.setPDFA3(true); // 1` works, too, while the second line is rendered useless by the `if (!documentPrepared)` condition in the `export` method, which seems to always fail for exporters made by the `A3Factory`.

Is there a real use case where the exporter's `prepareDocument` method does not need to be called?

jstaerk:
e.g. in case of xml export, both for FacturX as well as XRechnung. I remember a PR addressing this topic, is it now solved?

PR #91, opened 2019-06-18
Fix Issue #90: PDF/A-3 conversion missing in ZUGFeRDExporter.export(O…

tarikosman:
Fix for Issue #90: PDF/A-3 conversion missing in ZUGFeRDExporter.export(OutputStream output)
https://github.com/ZUGFeRD/mustangproject/issues/90

I added a PDF/A-3 check function to `MustangReaderWriterTest` and created `testMigratePDFA1ToA3Stream` to test the change. Additionally I called the PDF/A-3 check from the existing `testMigratePDFA1ToA3` which already converts to PDF/A-3.


jstaerk:
thanks

PR #92, opened 2019-06-22
Byte order marks cause unit test failures with Java 1.8

swsch:
I stumbled on this while looking for something else.

Also: MustangReaderWriterTest.java references a resource `/MustangGnuaccountingBeispielRE-20190610_507blanko.pdf`, which is not in the repository.

### DIFF
--- a/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDImporter.java
+++ b/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDImporter.java
@@ -369,7 +369,7 @@ public class ZUGFeRDImporter {
 	static String convertStreamToString(java.io.InputStream is) {
 		// source https://stackoverflow.com/questions/309424/how-do-i-read-convert-an-inputstream-into-a-string-in-java referring to
 		// https://community.oracle.com/blogs/pat/2004/10/23/stupid-scanner-tricks
-		Scanner s = new Scanner(is).useDelimiter("\\A");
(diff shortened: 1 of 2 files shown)


jstaerk:
I agree to the first change, thanks, but I would prefer to leave the BOM in the test case for custom XML because that is a) valid and b) a test if it is later removed and parsed correctly

swsch:
A BOM makes sense for byte oriented things like input streams or byte arrays. A string as a sequence of chars already is encoded and hence BOMs are not required. At least that's my understanding.

So if that test used a `byte[]`, including BOMs would be a great idea.

With the BOM, the unit test complains about it:
```
testCustomZF2Export(org.mustangproject.ZUGFeRD.MustangReaderWriterCustomXMLTest)  Time elapsed: 0.097 sec  <<< ERROR!
org.mustangproject.ZUGFeRD.ZUGFeRDExportException: org.xml.sax.SAXParseException; lineNumber: 1; columnNumber: 1; Content ist nicht zulässig in Prolog.
	at org.mustangproject.ZUGFeRD.ZUGFeRDImporter.extractString(ZUGFeRDImporter.java:200)
```

jstaerk:
the bom does (in the beginning of an XML file) does appear in the wild (see corpus, e.g. some french official examples) and i want that testet. Anyway I might add another test.

swsch:
oh dear ... I expected too much of the java input stream handling implementation, my bad. The BOM is not discarded but turns up as first codepoint in the string, which then needs to be discarded before the XML parser gets its hand on the data.

Issue #95, opened 2019-07-01
ram:CategoryCode 380 appears in ram:ApplicableTradeTax


swsch:
The [validation portal](https://www.din-zugferd-validation.org/validationportal/) complains about an invalid CategoryCode in the XML (v1) generated by mustang. There is a *380* in the trade tax total for the invoice, while the CategoryCode is *S* for individual trade tax items. 

```<rsm:CrossIndustryDocument ...>
    <rsm:SpecifiedExchangedDocumentContext>...</rsm:SpecifiedExchangedDocumentContext>
    <rsm:HeaderExchangedDocument>...</rsm:HeaderExchangedDocument>
    <rsm:SpecifiedSupplyChainTradeTransaction>
        <ram:ApplicableSupplyChainTradeAgreement>...</ram:ApplicableSupplyChainTradeAgreement>
        <ram:ApplicableSupplyChainTradeDelivery>...</ram:ApplicableSupplyChainTradeDelivery>
        <ram:ApplicableSupplyChainTradeSettlement>...
            <ram:ApplicableTradeTax>
                <ram:CalculatedAmount currencyID="EUR">68.99</ram:CalculatedAmount>
                <ram:TypeCode>VAT</ram:TypeCode>
                <ram:BasisAmount currencyID="EUR">363.10</ram:BasisAmount>
 ***            <ram:CategoryCode>380</ram:CategoryCode>
                <ram:ApplicablePercent>19.00</ram:ApplicablePercent>
            </ram:ApplicableTradeTax>...
        </ram:ApplicableSupplyChainTradeSettlement>
        <ram:IncludedSupplyChainTradeLineItem>
            <ram:AssociatedDocumentLineDocument>...</ram:AssociatedDocumentLineDocument>
            <ram:SpecifiedSupplyChainTradeAgreement>...</ram:SpecifiedSupplyChainTradeAgreement>
            <ram:SpecifiedSupplyChainTradeDelivery>...</ram:SpecifiedSupplyChainTradeDelivery>
            <ram:SpecifiedSupplyChainTradeSettlement>
                <ram:ApplicableTradeTax>
                    <ram:TypeCode>VAT</ram:TypeCode>
 ***                <ram:CategoryCode>S</ram:CategoryCode>
                    <ram:ApplicablePercent>19.00</ram:ApplicablePercent>
                </ram:ApplicableTradeTax>...
            </ram:SpecifiedSupplyChainTradeSettlement>
```

[...]

PR #97, opened 2019-07-03
deprecate misleading documentCode in favour of categoryCode

swsch:
pull request for #95.

passes tests and produces expected category code in XML file


jstaerk:
thanks a million!

Issue #23, opened 2017-05-25
have unit tests for nDigitFormat


jstaerk:
test cases are already documented in the comments of the function :-)

swsch:
I'm confused about nDigitFormat in class ZUGFeRDTransactionModelConverter, as it takes a BigDecimal argument, returns a BigDecimal argument, but still takes a round-trip through String-land.

Shouldn't 
```
	private BigDecimal nDigitFormat(BigDecimal value, int scale) {
		return value.setScale(scale, BigDecimal.ROUND_UP);
	}
```
be enough in this situation?

swsch:
The nDigitFormat method in class ZUGFeRD2PullProvider seems to do what [BigDecimal.toPlainString](https://docs.oracle.com/javase/8/docs/api/java/math/BigDecimal.html#toPlainString--) does.

That would make this method a simple wrapper:
```
	private String nDigitFormat(BigDecimal value, int scale) {
		return value.setScale(scale, BigDecimal.ROUND_UP).toPlainString();
	}
```

`mvn test` succeeds with these implementation, so I'll prepare a PR for review.

jstaerk:
> 
> ```
> 	private BigDecimal nDigitFormat(BigDecimal value, int scale) {
> 		return value.setScale(scale, BigDecimal.ROUND_UP);
> 	}
> ```
> 
> be enough in this situation?

should probably be ROUND_HALF_UP but yes, sorry, it's confusing, I removed it. Without unit tests in the first place which were only for the string returning ndigitformat (crossing fingers)

Issue #105, opened 2019-07-19
does not build on windows


jstaerk:
...allegedly...

swsch:
The alledged problem is probably our [old friend](https://github.com/ZUGFeRD/mustangproject/blob/24b27a9b516df3bf753119a31f0788c23b990396/src/test/java/org/mustangproject/ZUGFeRD/MustangReaderWriterCustomXMLTest.java#L67) where `new String(UTF8ByteOrderMark)` uses platform specific encodings. Either provide an explicit encoding as second argument to the string constructor or prefix the string with the character corresponding to the byte oder mark, see [this comment](https://github.com/ZUGFeRD/mustangproject/issues/99#issuecomment-511134151) and the next one.

Basically this is the same as #99

swsch:
d47f85c should fix this issue.

Not sure why it pops up in the pull request for #23. Probably due to lack of sleep last night.

jstaerk:
great! thanks a lot!

PR #106, opened 2019-07-21
simplify nDigitFormat implementations

swsch:
PR for #23


jstaerk:
I think I had tried that, I will implement #23 and see if it still works. Will take a while. Besides: could your unicode be #105 ?

jstaerk:
unfortunately it fails the new unit test in BaseTest, potentially only because of round_up instead of round_half_up  but to be honest I don't want to risk that because I am also not sure that toPlainString() does not apply locale settings, which it absolutely should not

PR #108, opened 2019-07-22
specify charset to avoid platform specific defaults

swsch:
This change fixes the build problem and lets `mvn test` succeed.


swsch:
... obviously not my best day interacting with github ...

jstaerk:
worked on it yesterday night and thought i would work on it tonight as well. Can not test it now but looks very promising, thanks!

PR #109, opened 2019-07-22
use RoundingMode enumeration and locale-safe BigDecimal.toPlainString

swsch:
I checked the implementation of BigDecimal.toPlainString in OpenJDK 1.8 (and 12) and as far as I can tell, the decimal point comes from the following helper method, which hardcodes the `.`:
```
    /* Returns a digit.digit string */
    private String getValueString(int signum, String intString, int scale) {
        /* Insert decimal point */
        StringBuilder buf;
        int insertionPoint = intString.length() - scale;
        if (insertionPoint == 0) {  /* Point goes right before intVal */
            return (signum<0 ? "-0." : "0.") + intString;
        } else if (insertionPoint > 0) { /* Point goes inside intVal */
            buf = new StringBuilder(intString);
            buf.insert(insertionPoint, '.');
            if (signum < 0)
                buf.insert(0, '-');
        } else { /* We must insert zeros between point and intVal */
            buf = new StringBuilder(3-insertionPoint + intString.length());
            buf.append(signum<0 ? "-0." : "0.");
            for (int i=0; i<-insertionPoint; i++)
                buf.append('0');
            buf.append(intString);
        }
        return buf.toString();
    }
```
Also, all tests pass.

### DIFF
--- a/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
+++ b/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
@@ -24,6 +24,7 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.io.UnsupportedEncodingException;
 import java.math.BigDecimal;
+import java.math.RoundingMode;
 import java.text.DecimalFormat;
 import java.text.DecimalFormatSymbols;
 import java.text.SimpleDateFormat;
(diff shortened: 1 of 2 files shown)


swsch:
re RoundingMode enumeration: This seems to be the type-safe successor to the "old-style" integer constants in BigDecimal, which are deprecated starting with Java 9.

jstaerk:
will have a look

jstaerk:
it passes the tests so i guess it works, thanks

Issue #73, opened 2019-03-16
Switch default ZF version


jstaerk:
currently version 1 is the default and version2 has to be enabled by default (I think it's sth like setZUGFeRDVersion()). In future, v2 should be default and v1 needs to be enabled by default.
How you can help:
This is a trivial change since I think it's the change of the value of is a single member variable.
Send me a pull  request. And by all means, change existing unit tests to factur-x and write unit tests for ZF2.

Issue #103, opened 2019-07-15
Thoughts on encoding XML entities


swsch:
Expanding my test cases ...

One of the products I found has a name like `Bestellerweiterung für E&F Umbau`. This confuses the XML parser when I'm attaching the ZUGFeRD file to the PDF:
```
org.dom4j.DocumentException: Error on line 26 of document  : Referenz zu Entity "F" muss mit dem Begrenzungszeichen ";" enden.
	at org.dom4j.io.SAXReader.read(SAXReader.java:462)
	at org.dom4j.DocumentHelper.parseText(DocumentHelper.java:286)
	at org.mustangproject.ZUGFeRD.ZUGFeRD2PullProvider.getXML(ZUGFeRD2PullProvider.java:143)
	at org.mustangproject.ZUGFeRD.ZUGFeRDExporter.PDFattachZugferdFile(ZUGFeRDExporter.java:375)
```

Is the user responsible for encoding the basic XML entities `<`, `>`, `&`, `"` and `'` in the input? If so, the documentation should say so more loudly.

If we forced encoding in the ZUGFeRD*PullProvider, it would break the code of users who already encoded the strings in their code.

In the interest of streamlining the user experience : How about providing a helper [like this one](https://stackoverflow.com/a/48588062) as, say, `ZUGFeRDExporter.encode(string)`, and introducing it in the example on the website with a company name like "Eine GmbH & Co. KG"? That would be in step 14 in "Write a ZUGFeRD-PDF file".

jstaerk:
no, that's a bug. I thought I had added a unit test with < but somehow that seemed to work. I added a unit test ZF2EdgeTest for this kind of cases but it does not yet fail, I guess it's soon time to merge ZUV (and integrate in the unit tests). The helper looks reasonable but it should automatically be called internally on all strings.

Despite this is clearly my fault because I still refuse to switch to JAXB for ZF2 (#87) despite it is already used for ZF1 (these kind of issues do not happen there):  I do still consider JAXB a big overhead.

PR #110, opened 2019-07-22
switch to ZUGFeRD v2 as DefaultZUGFeRDVersion

swsch:
this is the trivial change for #73


jstaerk:
thanks, I'd like to do #103 first and I'm not sure if in the commandline (called toecount) we don't need to change a filename but I'll keep that merge request for later :-)

jstaerk:
now merged into alpha version of mustang 2

Issue #113, opened 2019-07-25
v1 documents ignore profile setting


swsch:
Looks like v1 ZUGFeRD documents always end up with the extended profile.
#112 is the first step to fix this.

PR #114, opened 2019-07-25
add failing unit test to prove that the wrong profile is written and then fix it

swsch:
re #113:

This test fails at line 411, which means that the XML file does not reference the BASIC profile, but does reference the EXTENDED profile and hence cannot reference the COMFORT profile as requested in line 400.


swsch:
commit 9681b4f extends the IProfileProvider interface with a method for setting the profile via the values of the ZUGFeRDConformanceLevel enumeration. 

This makes the failing unit test work for v1 and does not break the unit test for v2 and custom xml.

Issue #118, opened 2019-07-31
access to TotalPrepaidAmount in v1


swsch:
This seemed to be quite similar to the recent issue #115, but it turns out to have further reaching effects, as the element's parent is `ram:SpecifiedTradeSettlementMonetarySummation` in v1 and `ram:SpecifiedTradeSettlementHeaderMonetarySummation` in v2.

jstaerk:
implementation in v2 should not be too complicated , maybe sth like three, four lines in ZUGFeRD2pullprovider:generateXML are enough

swsch:
good estimation, see #120

swsch:
guess what ... three lines and I managed to mess up one of them :-)

PR #119, opened 2019-07-31
Implement TotalPrepaidAmount

swsch:
This is a manually tested implementation for #118.

Updating the existing tests seems to be ... troublesome.


jstaerk:
you could add a prepaidamount in the mustangreaderwriteredgetest

swsch:
That is, in fact, what I start with yesterday. Then I stumbled over the difference in element names, and somehow messed up thoroughly. So, after a `git reset --hard` I went a tad more slowly. :-)

PR #123, opened 2019-08-05
parse rawXML into document only once

swsch:
I noticed that `getDocument` parses the contents of `rawXML` for each invocation of `extractString`.

Here I'm using a setter for rawXML to parse `rawXML` once after changing its value so that `getConfig` can now return this prepared Document. Current unit tests pass and even run a tiiiiny bit faster than before.


jstaerk:
thanks a lot!

PR #130, opened 2019-08-12
IZUGFeRDExportableContact to default-Methods; added ShipToTradeParty and BuyerOrderReferencedDocument

pierrebarke:
switched IZUGFeRDExportableContact to default-Methods; added
ShipToTradeParty: if the merchandies is not shipped to the same address
as stated in the invoide; added BuyerOrderReferencedDocument: the ID of
the purchase order the invoice ist based on


jstaerk:
thanks a lot

PR #131, opened 2019-08-12
removed unsed Methode; IssuedDateTime only if not null

pierrebarke:
removed @getBuyerOrderReferencedDocument: not used after merge; only call
@setIssueDateTime if not null to avoid tag with NULL value

Thanks for merging my changes. i did some changes after merge for clean up


jstaerk:
thanks

Issue #136, opened 2019-10-11
Support different Date formats


Koizumi85:
Hello everyone,

following the ZUGFeRD 2 specification, it should be possible to specify some dates in the following formats:
- month of year, format `yyyyMM` with code `610`
- week of year, format `yyyyww` with code `616`
- day, format `yyyyMMdd` with code `102`

At the moment as far I can see, mustangproject does only support the last one.
But especially for the delivery date, the possibility to specify the other two formats, would be absolutely essential.

I would suggest to provide a new interface `IZUGFeRDDate` which has the following methods
`java.util.date getDate()`
`ZUGFeRDDateFormat getFormat()`
where `ZUGFeRDDateFormat` is an enum providing the following information for each of the formats:
- An instance of SimpleDateFormat, which can be used to format the given date
- The ZUGFeRD code which belongs to this date format

With this information it would be possible to write the generation of the "occurence date time" section to something like the following:

```
if (trans.getDeliveryDate() != null) {
   // same like before
} else if (trans.getZUGFeRDDeliveryDate() != null && trans.getZUGFeRDDeliveryDate().getDate() != null) {
   ZUGFeRDDateFormat dateFormat = trans.getZUGFeRDDeliveryDate().getFormat();
   Date date = trans.getZUGFeRDDeliveryDate().getDate();
   xml += "<udt:DateTimeString format=\"" + dateFormat.getFormatCode() + "\">"
+ dateFormat.getFormatter().format(date) + "</udt:DateTimeString>";
} else {
  throw new IllegalStateException("no delivery date provided");
}
```

this pattern could then also be used for other dates, where it makes sense. What do you think?

jstaerk:
Hi


> pattern could then also be used for other dates, where it makes sense. What do you think?
It's only important that it needs to stay compatible to the zf1 interface, which has been done with JAXB, and with the current implementation. If that is the case I would very much appreciate according contributions.

kind regards
Jochen

Koizumi85:
Okay. I will try to provide a pull request for it.

Koizumi85:
closed by #137

PR #137, opened 2019-10-14
deliveryDate in different date formats (#136)

Koizumi85:
This pull request addresses issue #136 by
- Added possibility to specify delivery date in "week of year" or "month of year" format, through introduction of new IZUGFeRDDate interface. Keeping backward compatibility by providing both methods in IZUGFeRDExportableTransaction

- Added possibility to specify "LineTwo" in Recipient Address

- Limited pdfbox version to maximum 2.0.17 because otherwise 3.0.0-SNAPSHOT was pulled in which was broken because of missing ByteArrayDataSource

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -79,12 +79,12 @@
 		<dependency>
 			<groupId>org.apache.pdfbox</groupId>
 			<artifactId>preflight</artifactId>
-			<version>[2.0.15,)</version>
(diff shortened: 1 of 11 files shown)


Koizumi85:
@jstaerk Thanks for merging :) What do you think when the next version including these features will be released?

PR #158, opened 2020-05-28
Add extraction methods

aberndt-hub:
Added functions for extracting LineItems and modified some existing extractions functions to match ZUGFeRD 1.x and ZUGFeRD 2.x


jstaerk:
thanks

PR #173, opened 2020-08-04
Proposal for loading resource - V1.7.8

Cakmakli-a:
`Validator`

```
        @Override
        protected void validateSchema(final byte[] xmlRawData, final String schemaPath, final int section, final EPart part) throws IrrecoverableValidationError {
            final URL schemaFile = Validator.class.getResource("/schema/" + schemaPath);
            final javax.xml.transform.Source xmlData = new StreamSource(new ByteArrayInputStream(xmlRawData));
            final SchemaFactory schemaFactory = SchemaFactory.newInstance("http://www.w3.org/2001/XMLSchema");

            try {
                final Schema schema = schemaFactory.newSchema(schemaFile);
                final javax.xml.validation.Validator validator = schema.newValidator();
                validator.validate(xmlData);
            } catch (SAXException var10) {
                context.addResultItem((new ValidationResultItem(ESeverity.error, "schema validation fails:" + var10)).setSection(section).setPart(part));
            } catch (IOException ioe) {
                LOGGER.error(ioe.getMessage(), ioe);
                throw new IllegalStateException("Validation fails and ends with IOException!");
            }
        }
```

First line in this method is important. It is a "safe alternative" instead of using the ClassLoader.

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,9 +1,16 @@
+1.7.8
+=====
+2020-06-14
+
+- \#159 gross price in ZF2 should be without VAT
+
 1.7.7
 =====
+2020-05-26
 
 - Refactored comparison operator for ChargeIndicator https://github.com/ZUGFeRD/mustangproject/pull/153
-- #148 support additional documents for items
-- #154 german VAT ID may not be used for sellercontact scheme id
+- \#148 support additional documents for items
+- \#154 german VAT ID may not be used for sellercontact scheme id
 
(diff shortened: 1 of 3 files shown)


Cakmakli-a:
What we are gonna do now with this proposal?

tklein:
Is someone already working on this? And/or what needs to be done to get this merged?

TIA

PR #176, opened 2020-09-19
ZUGFeRD2PullProvider: Add write support for GlobalID and ID to buyer addresses

heisej:
especially for 'BuyerTradeParty' and 'ShipToTradeParty' right now.

Since EN16931 allows only one ID entry for these entries, the ID assigned
by the buyer is prefered over the global ID for now if bot are defined in
the IZUGFeRDExportableTradeParty object.


jstaerk:
thank you

Issue #186, opened 2020-11-10
Wrong categoryCode in <ram:ApplicableHeaderTradeSettlement>


weclapp-zugferd:
When there are several entries of <ram:SpecifiedLineTradeSettlement> with different CategoryCode (e.g. "S" for nominal tax and "Z" for tax free) then the summary in <ram:ApplicableHeaderTradeSettlement> contains only categoryCode "S". This results in validationErrors.

![xmlSnippet](https://user-images.githubusercontent.com/74199346/98644415-ba8ae800-2330-11eb-96bc-d6e686beddf7.png)
![xmlSnippetWithError](https://user-images.githubusercontent.com/74199346/98644418-bb237e80-2330-11eb-8aee-83337f6e0e87.png)
![ValidationWithErrors](https://user-images.githubusercontent.com/74199346/98644411-b959bb00-2330-11eb-95ef-97c66d4de8ae.png)

jstaerk:
was corrected with https://github.com/ZUGFeRD/mustangproject/pull/187 , thank you

PR #187, opened 2020-11-10
Correct tax category code

weclapp-zugferd:
Use TaxCategoryCode instead of DocumentCode by creating VATPercentAmountMap. #186


jstaerk:
looks good, thanks!

Issue #185, opened 2020-11-09
original jars to Maven Central


quadrik:
Please provide also the original (non-shaded) jars of library and validator via Maven Central, maybe using a classifier: 'org.mustangproject:validator:2.0.0-alpha3:original' . Those jars can be included in own projects without colliding with other 3rd party dependencies.

jstaerk:
I would love to, can you tell me how? I more or less only followed https://dzone.com/articles/publish-your-artifacts-to-maven-central and I had a look, I don't immediately see how to publish the nonshaded jars

jstaerk:
@quadrik I have an issue merging #214 because it contradicts your solution. So when I make #213 work and upload everything the non-shaded artefacts would no longer be available via maven central?

quadrik:
yes, the non-shaded jar for the library won't be published. I can reproduce your error locally. Although the resulting shaded CLI.jar contains the class XmlJavaTypeAdapter an exception is thrown. If you compare the CLI jars (shadedArtifactAttached false vs true), the latter is smaller and lacks a bunch of classes, but I don't know exactly why. Maybe you need to define additional dependencies for the CLI module.

quadrik:
the actual exception was swallowed and goes like this:
```
java.lang.reflect.InvocationTargetException
	at jdk.internal.reflect.GeneratedMethodAccessor2.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:567)
	at com.sun.xml.bind.v2.model.annotation.RuntimeInlineAnnotationReader.getClassValue(RuntimeInlineAnnotationReader.java:107)
	at com.sun.xml.bind.v2.model.annotation.RuntimeInlineAnnotationReader.getClassValue(RuntimeInlineAnnotationReader.java:27)
	at com.sun.xml.bind.v2.model.impl.PropertyInfoImpl.isApplicable(PropertyInfoImpl.java:186)
	at com.sun.xml.bind.v2.model.impl.PropertyInfoImpl.getApplicableAdapter(PropertyInfoImpl.java:197)
	at com.sun.xml.bind.v2.model.impl.PropertyInfoImpl.<init>(PropertyInfoImpl.java:96)
[...]

PR #189, opened 2020-11-10
have shaded artifacts attached

quadrik:
#185: shaded jars are attached and published with classifier 'shaded' and no longer replace the original jars


jstaerk:
ok lets try. Next release will be the 2.0 and if that one breaks it I will be angry.

jstaerk:
that actually worked, thanks a lot. I have a question, how can i prevent "core" to be published?(the people need to use at least library instead...)

quadrik:
Was a pleasure. As "core" is your parent pom the other modules depend upon, you need to deploy it too, I guess. It's the same for other open source projects, if you take a look .

Issue #191, opened 2020-11-16
SpecifiedTradeAllowanceCharge with same taxes are enumerated instead of summed


weclapp-dev:
If there are shippingCosts and document specific discount then the result is invalidDoc.pdf (pdf errors can be ignored). The right document should look like validDoc.pdf.
[validDoc.pdf](https://github.com/ZUGFeRD/mustangproject/files/5547043/validDoc.pdf)
[invalidDoc.pdf](https://github.com/ZUGFeRD/mustangproject/files/5547044/invalidDoc.pdf)

jstaerk:
since PR 192 has been merged this is now solved?

weclapp-dev:
Yes, it is solved. Thank you.

PR #192, opened 2020-11-16
BigDecimal specific refactoring

weclapp-dev:
- Use  BigDecimal.stripTrailingZeros() as key for getVATPercentMap and in consequence use compareTo... == 0 instead of equals
- Use BigDecimal.ZERO instead of new BigDecimal(0)
- These changes result in improvements in enumeration of the <ram:SpecifiedTradeAllowanceCharge> tags and the calculation of tag <ram:SpecifiedTradeSettlementHeaderMonetarySummation>. #191

### DIFF
--- a/library/src/main/java/org/mustangproject/Charge.java
+++ b/library/src/main/java/org/mustangproject/Charge.java
@@ -16,7 +16,7 @@ public class Charge implements IZUGFeRDAllowanceCharge {
 	protected String categoryCode;
 
 	public Charge() {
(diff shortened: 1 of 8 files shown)


jstaerk:
Thanks a lot. Besides: if you belong to the weclapp company, they list Neue Mainzer Straße in Frankfurt as one of the addresses. I'm there as well, on tuesdays and thursdays I'm in Mindspace, room 313, if you like drop me a visit and we can have a coffee together.

PR #195, opened 2020-11-18
Improve VATAmount calculation

weclapp-dev:
I accidentally found a constellation of items and allowances where the previous implementation of LineCalculator calculated total VAT amount wrong (1ct too much).
A test in the test class shows the example.


jstaerk:
Thanks a lot.

FYI instead of IZUGFeRDAllowanceChargeImpl() and setItemAllowances you could have also used Charge()/Allowance() and addCharge()/addAllowance() as in https://github.com/ZUGFeRD/mustangproject/blob/master/library/src/test/java/org/mustangproject/ZUGFeRD/ZF2PushTest.java:testItemChargesAllowancesExport, or set the price to  163,603 in the first place

PR #202, opened 2020-11-25
XRechnung importer

weclapp-dev:
Just another importer to import an XRechnung in xml


jstaerk:
fundamentally OK but maybe somewhen we should maybe call it CIIimporter instead.
And maybe also ZUGFeRDInvoiceImporter -> CIIInvoiceImporter https://www.mustangproject.org/invoice-class/?lang=de#reading

PR #204, opened 2020-11-30
Update sample MustangWriter for Mustangproject 2.0.2

tweimer:
This fixes the compile errors that came up after trying to use the sample file.
Now it compiles fine and produces ZF2

Most likely, other changes have to be done.


jstaerk:
thank you so much, I was in the process of updating the docs and that was on my list but I just didnt manage so far

PR #207, opened 2020-12-04
Improvements

weclapp-dev:
- Import UnitCode of basisQuantity to product.
- Improve handling of categoryCode
- Fix id of profile XRECHNUNG


jstaerk:
thanks a lot but I will revert the XRechnung version back to 1.2: unfortunately we will have to solve differently since currently 2.0 is still rejected from ZRE (and I have no idea how they intend to transition)

PR #218, opened 2021-01-27
Allow setting Contact for recipient TradeParty in XRechnung

seeeeew:
This adds support for setting a Contact for the recipient (buyer) TradeParty while using the XRechnung profile in accordance with the format specifications. I've tested this in all relevant combinations and validated the results with the Mustang validator without errors.

Please tell me if anything else is required to get this merged, since I didn't find any contribution guidelines.

### DIFF
--- a/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
+++ b/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
@@ -129,7 +129,7 @@ public class ZUGFeRD2PullProvider implements IXMLProvider {
 		}
 		xml += "	<ram:Name>" + XMLTools.encodeXML(party.getName()) + "</ram:Name>\n";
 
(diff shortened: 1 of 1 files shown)


jstaerk:
thanks, sorry for overlooking

jstaerk:
...wait. If it's disallowed in EN16931 it can't be allowed in a subset?

seeeeew:
I'm not sure what you mean. I have to admit I'm not an expert on EN16931, I just checked the XRechnung 2.0.1 specification i found [on the KoSIT website](https://www.xoev.de/xrechnung-16828#DL) (e. g. 11.3 in `201-XRechnung-2020-12-16.pdf`). [Their test suite](https://github.com/itplr-kosit/xrechnung-testsuite) also contains lots of cases where the BuyerTradeParty has a DefinedTradeContact.

jstaerk:
You are right, I checked, contact is apparently only forbidden for "end" recipient and "goods" recipient but not for (invoice) recipient, I should fix that and also allow a contact for (invoice) recipients in en16931

seeeeew:
Ah good, for a second I was worried I made a mistake. Thank you!

PR #219, opened 2021-01-27
activate shade-plugin only for module CLI; fixes ClassNotFoundExcepti…

quadrik:
1. fixes the ClassNotFoundException mentioned in #214; see comment in #185
2. I activated the shade-plugin only for module CLI, because in my opinion you don't really need shaded jars for library and validator. If you use library/validator in your own project (like me), Maven/Gradle will pull in the other jars transitively.

### DIFF
--- a/Mustang-CLI/pom.xml
+++ b/Mustang-CLI/pom.xml
@@ -37,6 +37,11 @@
         -->
       </dependency>
       <dependency>
+         <groupId>com.helger</groupId>
+         <artifactId>en16931-cii2ubl</artifactId>
+         <version>1.3.0</version>
+      </dependency>
+      <dependency>
          <!-- for commandline options -->
          <groupId>com.sanityinc</groupId>
          <artifactId>jargs</artifactId>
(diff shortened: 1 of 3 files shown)


jstaerk:
the background for reverting the change was that cii2ubl did not work with it, does `--action ubl` still generate a UBL from a CII file after your patch? and does the `<classifier>shaded</classifier>` help/do you need it (RE https://groups.google.com/g/zugferd/c/Riml-eJmH70 )

quadrik:
yes I tested the conversion via CLI.jar with an example-CII-file and it worked. With my patch the shaded-classifier wouldn't be available anymore. But you can ignore this PR and just add the new helger-dependency to module CLI and of course reset 'shadedArtifactAttached' to true

jstaerk:
sorry but that adds en16931-cii2ubl as a dependency to the CLI, but it is functionality that should be provided by the library. So in this case I do need a shaded library (and in case the user does not yet use PDFbox, saxon) which you might be able to reference with classifiers but I don't want to tell my users if they need ubl functionality they also manually have to include another jar :-(

PR #240, opened 2021-04-15
Set ignoreInputErrors flag before loading PDFs

fourcube:
I noticed that PDF files are still checked for PDF/A conformity even though the  `-i` was set. This fix makes the flag work as it should (at least from my expectations).


jstaerk:
looks good, thank you! Could you have a look if the current dev version works for you if you `mvn package`?

jstaerk:
i tested myself, it works, thanks again

fourcube:
Thank you for maintaining this great project and getting this merged so quickly!

PR #241, opened 2021-04-15
Use typecode 58 (SEPA) in SpecifiedTradeSettlementPaymentMeans

heisej:
as old code 42 is for the more generic "payment to bank account" while
the ZUGFeRD specification and EN16931-1 recommend to clearly distinguish
between SEPA transfers and non SEPA transfers and we are only supporting
SEPA payment data right now.


jstaerk:
thanks a lot!

PR #257, opened 2021-12-17
solves #227

quadrik:
* as for module validation, creates an uber jar with classifier "shaded" besides the simple non-shaded jar
* the integration test for CLI assures that Cii2Ubl conversion is still working


quadrik:
Hi Jochen, with 86f3aee81628492a7b7f you reverted my changes to library/pom.xml . Can you set shadedArtifactAttached to true ? Regards

Issue #270, opened 2022-03-31
Contrary to what the the prompt message states, an "Attachment filename" must be specified 


AlexGeller1:
```shell
$java -jar /home/alex/ZUGFeRD/mustangproject/Mustang-CLI/target/Mustang-CLI-2.4.1-SNAPSHOT.jar --action=combine --format=zf --version=2 --profile=M --source Avoir_FR_type380_MINIMUM.pdf --source-xml=factur-x.xml --out Avoir_FR_type380_MINIMUM.ZUGFeRD.pdf
12:50:33,757 |-INFO in ch.qos.logback.classic.LoggerContext[default] - Could NOT find resource [logback-test.xml]
...
Source PDF set to Avoir_FR_type380_MINIMUM.pdf
ZUGFeRD XML set to factur-x.xml
Output PDF set to Avoir_FR_type380_MINIMUM.ZUGFeRD.pdf
Attachment filename (empty for none) (default: ):
File does not exist, try again or CTRL+C to cancel
Attachment filename (empty for none) (default: ):
```

jstaerk:
I believe you might be testing master head and not a released version but actually you are right, this needs to be addressed before the next version can be released: please help!

jstaerk:
thanks for your PR, it's merged and it works

PR #271, opened 2022-04-01
Fixed issue that specifying an attachment was mandatory (issue #270)

AlexGeller1:
Also in this PR:

- Making the --attachment option operational
- Renaming --attachment to --attachments
- Adding the "ox" format to the usage output

Please note that specifying an empty string for the --attachments option is the way to prevent the program from prompting for attachments. That perhaps is a bit clunky but I don't have a better idea. 

Something else, I noticed lines like this:
```Java
Boolean filesFromStdIn = cmd.hasOption("listfromstdin");
```
The signature of the called function however is (See here: https://commons.apache.org/proper/commons-cli/javadocs/api-release/org/apache/commons/cli/CommandLine.html#hasOption-java.lang.String-):
```Java
public boolean hasOption(String)
```
Any reason for this?

### DIFF
--- a/Mustang-CLI/src/main/java/org/mustangproject/commandline/Main.java
+++ b/Mustang-CLI/src/main/java/org/mustangproject/commandline/Main.java
@@ -72,11 +72,12 @@ public class Main {
 				+ "                [--source=<filename>]: set input PDF file\n"
 				+ "                [--source-xml=<filename>]: set input XML file\n"
 				+ "                [--out=<filename>]: set output PDF file\n"
-				+ "                [--format <fx|zf>]: set ZUGFeRD or FacturX\n"
+				+ "                [--format <fx|zf|ox>]: set Factur-X, ZUGFeRD or Order-X\n"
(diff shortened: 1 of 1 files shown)


jstaerk:
coolio, thanks a lot, looks good at first sign, might have a closer look at the WE

jstaerk:
> 
> Something else, I noticed lines like this:
> 
> ```java
> Boolean filesFromStdIn = cmd.hasOption("listfromstdin");
> ```
> Any reason for this?

you mean it should become 
```java
 boolean filesFromStdIn = cmd.hasOption("listfromstdin");
```
? 
You are probably right, the reason was the (attempted?) switch from JArgs to 
Apache Commons Cli, the reason for the switch being a) JArgs failing to publish a non-snapshot version for quite a while now (which means that whenever Mustang is released I have to explicitly confirm to publish using a project that has no officially usable version) and b) to be able to specify attachments as argument I need a way to also understand arguments for multiple attachments.

Issue #262, opened 2022-02-01
Bundled AssertJ shadows current version


florianklumb:
Hi Jochen,

We're implementing a web app for a German government organization and now have a CR to implement an XRechnung based invoicing. Unfortunately, Mustang seems to be bundled with a certain flavor and version of AssertJ that shadows the version we're using in our project (`org.assertj.core.api`). How do I notice? For instance, the bundled AssertJ seems to not include `assertThatCode`.

Have you encountered this problem before? Do you have a hint what we can do?

Thanks

Florian

florianklumb:
OK I figured it out: one just needs to explicitly include a current version of AssertJ:
```
    implementation "org.assertj:assertj-core:3.22.0"
    implementation "org.mustangproject:library:2.4.0"
```

jstaerk:
Hi,
1) Which version of AssetJ do you use?
2) how do you refer to Mustang? 
a) If you are using jars: Apart from http://mustangproject.org/deploy/library-2.4.0.jar there is also a original-library-2.4.0.jar and additionally to validator-2.4.0-shaded.jar there is also a unshaded validator-2.4.0.jar .
b) If you are using maven central: There were discussions about dependencies in (#185 and ) #227 before they were able to patch it, my point was (and still is) that fundamentally you should be able to use the unshaded artifacts if you take care about the dependencies yourself, as far as I understood shaded used to be a classifier, i.e. I believe the idea was that you can refer to <artifact id>:shaded for shaded assets and only to <artifact id> for an unshaded version.
kind regards
Jochen

florianklumb:
Hi Jochen,

Thanks for your reply. Unfortunately, shading was disabled again (see my reply in #227). I have to say that I cannot follow what you describe. Can you explain? 

Thanks

Florian

jarhantimwi:
Hello,

I have the exact same problem here.
Is there a way to get the slim jar from maven-central?

Thanks

Jacques

DFenzl:
Hello,

we also have the same problem here. We would prefer to get a slim version of mustang and handle the dependencies by our selfs, cause we have a very complex structure of sub projects with many other dependencies.

Thanks 

Dominik

jarhantimwi:
Hi,
I think the jar in maven central should be the one without all the dependencies and if you want to use the uber jar, you can make it available for download on the mustang website.
[...]

PR #274, opened 2022-06-15
issue 262 - bundle library jar without all dependencies

jarhantimwi:
Allow to bundle the library jar without all the dependencies and to use it in another application.
see https://github.com/ZUGFeRD/mustangproject/issues/262#issuecomment-1110062727


jstaerk:
Just a small question, if I merge this, how *do* I get a shaded version?

PR #288, opened 2022-10-07
Proposed changes of namespace and meta-data for Delivery-X

ivaklinov:
As of late 2022 the Delivery-X standard is not yet published. See specification:
Die digitale Ablösung des Papier-Lieferscheins, Version 1.1, April 2022 
Chapter 7.1 XMP-Erweiterungsschema für PDF/A-3
http://docplayer.org/230301085-Der-digitale-lieferschein-dls.html

Seeburger intends to deliver this as an early-access implementation, subject to change (when the standard is made official).


ivaklinov:
BTW: When trying to build the whole project (not just the Mustang library) from the current master branch, in its latest version, the maven build with JDK 11 on Ubuntu fails for the Validator project...

image

... not a problem for our use case. The validation seems to fail on file: `../library/target/testout-OX-Edge.pdf`

PR #292, opened 2022-10-11
ZUGFeRD 1 Import

StefanSchmaltz:
This pull request contains some adjustments for importing ZUGFeRD 1 invoices.

### DIFF
--- a/library/src/main/java/org/mustangproject/Invoice.java
+++ b/library/src/main/java/org/mustangproject/Invoice.java
@@ -20,14 +20,20 @@
  */
 package org.mustangproject;
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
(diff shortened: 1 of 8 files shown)


jstaerk:
Thanks for the PR. I added a unit test and parsing the invoice number and due date for ZF1 invoices (using sth like xpath.compile(`"//*[local-name()=\"ApplicableHeaderTradeSettlement\"]|//*[local-name()=\"ApplicableSupplyChainTradeSettlement\"]");`). 

However, I did remove grand total, due payable and total prepaid from the invoice class again, sorry, but that's task of the invoicecalculator. If I saw it correctly it was anyway being written with the calculated and not the read values. 

It makes sense to parse the prepaid amount and store it somewhere since in an imported invoice that's actually part of the payload but as long as nobody needs it I'll probably just keep it the way it is.

PR #302, opened 2022-12-07
Prevent NPE

weclapp-dev:
Maybe a null check before calling this method is the better way to prevent this exception? But in my opinion it is good enough.


jstaerk:
thank you, good enough and already better than my code

Issue #308, opened 2023-02-08
Give possibility to define when VAT becomes applicable (BT-8)


chris-webert-its4u:
Hello,

We're using Mustang to generate Factur-X invoices for our french suppliers so that they can import these in Chorus Pro directly. It works fine so far but a few of those suppliers work differently from the others, and need the VAT applicability (BT-8) on their invoices to be set at payment date rather than invoice issue date (the default). Chorus Pro let you choose that setting when you create an invoice through their portal, so it should be feasible to read it from a Factur-X at import I'd assume.

I found that you can do that with a Time code (UNTDID 2475) at the `ram:ApplicableTradeTax` level, with a tag called `ram:DueDateTypeCode`. Authorized values seems to be 5, 29 and 72 for UN/CEFACT CII.

I don't think there's a way to set that code with Mustang yet (or I may have missed it), do you think it's possible to add this in the future?

Thank you

jstaerk:
It is possible, we can do it together, you could pay for it (this sounds like a 3 digit euro effort) or you might have to wait until someone had the time. For my reference: e.g. in the FNFE Factur-X sample Facture_F20220023-LE_FOURNISSEUR-POUR-LE_CLIENT_EN_16931 there is a DueDateTypeCode of 72. The TA says the values mean 
- 5: date of the invoice (VAT on debits) - 29: delivery date (VAT on debits) - 72: payment date (VAT on income).

PR #309, opened 2023-03-03
Issue 308: define when VAT collection become applicable

chris-webert-its4u:
PR for issue #308 

Hi, I had some spare time to work on the VAT due date collection (see issue). It works fine for my use case but I'm not sure I have the necessary broad vision of the project to say it will be the case for everyone. Any input on this problem is very much welcomed :)


jstaerk:
1) thanks
2) Raymond says by opensource is "scratching a developer's personal itch" and Adenauer said "those who have visions should see the doctor". It's not that I would oppose a more suitable long term approach if there was any but so far nobody needed any feature in that domain, you are the first one, I have no idea so lets go with a working solution and refactor if it turns out there could be a cleaner more generic approach.

PR #313, opened 2023-04-18
Update helger library

weclapp-dev:
There is already version 2.0.1. Unfortunately the update results in many compatibility test failures and ClassNotFoundExceptions with to transitive dependencies. Therefore the step to 1.4.10 is sufficient for now.

Now without test failures.


jstaerk:
thanks a lot!

PR #324, opened 2023-06-07
323 - Maven Artifacts Naming too generic

jpep-in:


jstaerk:
that's a bigger change (all maven central users would have to adept) and I'm not precisely sure if it's my fault (artifact IDs should be arbitrary within a group) and if at all I would postpone to mustang 3. RE maven central artefacts my personal itch is currently that despite mutliple attempts I could never cease to release a empty "core" shell which is just the parent project of library and validator ;-(

jpep-in:
Hello Jochen,
Thank you for your answer.
I'm conscient that is major change, not possible to merge before a next major release.
In a multi module maven project with parent and childrens, it's not mandatory, but a best practice is to reuse parent's names in the artefactid
exemple: https://github.com/coldserenity/maven-naming-conventions

Naming is one of first thing seen, when a developer do his 'sanity check' to adopt or not a library. :)
I think for the adoption of mustang, this "detail" is also important.

I understand better now the name 'core', it's the parent project, so, hierarchy could had be :

mustang-parent
 ├─ mustang-library
 ├─ mustang-validator
 ├─ mustang-cli

I precise i'm not a maven top expert, but i have created or contributed to some multi modules projects.
If you decide to rename or re-organise project in the futur, an open discussion about could be interressant. :)

Best,

Issue #325, opened 2023-06-07
PdfA conformance validation utility with errors result


jpep-in:
Before export, it would be easier to execute just a simple validation and get the potential errors, than just exception and no explicit error during export pass process.

jpep-in:
PR #326

jstaerk:
please note the PDFbox quick check (is usually correct but) does not completely overlap with veraPDF which can be quite confusing but there is an option to turn the exception off if you explictly know what you are doing (ref #164 and #244)

Issue #244, opened 2021-06-22
pdf/A-1b falsely invalid ?


obonnefon:
Hello, here's a file considered by VeraPdf as a valid Pdf/A-1 but rejected by Mustang when trying to combine with xml file. 
(a pb of embedded font - Calibri)
juin 22, 2021 8:23:07 AM org.apache.pdfbox.pdmodel.font.PDFont loadUnicodeCmap
AVERTISSEMENT: Invalid ToUnicode CMap in font SVFUWB+Calibri,Bold
juin 22, 2021 8:23:07 AM org.apache.pdfbox.pdmodel.font.PDFont loadUnicodeCmap
AVERTISSEMENT: Invalid ToUnicode CMap in font CUMZOI+Calibri
juin 22, 2021 8:23:07 AM org.apache.pdfbox.pdmodel.font.PDFont loadUnicodeCmap
AVERTISSEMENT: Invalid ToUnicode CMap in font SVFUWB+Calibri,Bold
juin 22, 2021 8:23:07 AM org.apache.pdfbox.pdmodel.font.PDFont loadUnicodeCmap
AVERTISSEMENT: Invalid ToUnicode CMap in font CUMZOI+Calibri
08:23:07.383 [main] ERROR o.mustangproject.validator.Validator - File is not a valid PDF/A input file
java.io.IOException: File is not a valid PDF/A input file
        at org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA1.ensurePDFIsValid(ZUGFeRDExporterFromA1.java:81)
        at org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3.load(ZUGFeRDExporterFromA3.java:121)
        at org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA1.load(ZUGFeRDExporterFromA1.java:92)
        at org.mustangproject.commandline.Main.performCombine(Main.java:648)
        at org.mustangproject.commandline.Main.main(Main.java:403)

[invoice.pdf](https://github.com/ZUGFeRD/mustangproject/files/6691669/invoice.pdf)

Thanks

jstaerk:
Hi,

amacado:
I came across this issue and what might be useful for someone else encountering this problem using the `CLI` version: You can use the equivalent cli argument `[-i, --ignorefileextension]` which "Check for all files (*.*) instead of PDF files only (*.pdf) in metrics, **ignore PDF/A input file errors in combine**".

[...]

PR #326, opened 2023-06-07
325 - extract pdfa validation to utility

jpep-in:


jstaerk:
the pdfa check of _pdfbox_ should anyway be removed, the guys from verapdf have invested months and did a real great work. I more and more consider letting go of the "library" and publishing only validator as of mustang 3 onwards which would also mean that this check could be removed which anyway causes a lot of confusion (#164, #244 and #325)

PR #331, opened 2023-07-24
Improvement of included notes

weclapp-dev:
There is the possibility now to add notes with specific subject code (the recommended ones from the specification).
Some refactorings were done


jstaerk:
cool, thanks a lot

PR #336, opened 2023-09-06
new parameter "disable-file-logging" added

Heavenfighter:
Added new parameter "disable-file-logging" to disable logging to file.


jstaerk:
Well it works and it's great and now I wonder why I never managed. Thank you 🥳 I might release it in a 2.8.0 still this month.

PR #338, opened 2023-09-27
Migration from pdfbox 2 to pdfbox 3

pokint:
We migrated from pdfbox 2 to 3 in our software recently and noticed,  that the mustang library broke in the process. Since I didn't want to roll the changes back in our software I decided to migrate pdfbox in your project, too.

Be aware, this PR has **breaking changes**. I had to replace all load() functions with one that accepts a "File" object, since the signature of the PreflightParser changed significantly (eg:`PreflightParser.validate(pdfFile).isValid()`). I don't know why the Parser is so much more flexible by using RandomAccessRead-classes, but the PreflightPrarser lacks such constructors.

Due to this change I had to change getResourceAsStream to getResource, which should be fine, since the tests don't run in a jar-file.

Additionally I had to modify some tests, because getDocumentInformation is mostly deprecated. The producer can't be set using it anymore:
image

See also: [PDFBox Migration Guide](https://pdfbox.apache.org/3.0/migration.html)

### DIFF
--- a/History.md
+++ b/History.md
@@ -1,5 +1,7 @@
 2.10.0
-- 
+=======
+2023-12-30
+
 - also accept pdf/a3 from inputstream
 - closes #354 factur-x 1 from commandline
 - support XR 3.0.1 (#343), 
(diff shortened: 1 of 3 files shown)


jstaerk:
Suggestion: -> mustang version 3 could get rid of the "library" only proving the "validator" and then also the preflight check could be tackled by verapdf which would be always available

PR #361, opened 2024-02-18
Some general improvements for the README.md.

reiz:
Here some changes to the README: 

- I moved the license section to the bottom, because that's the best practice. GitHub is anyway parsing the license information directly from the pom.xml file and displays the license SPDX identifier on the right upper corner. That's why the license section is anyway somehow redundant on the README. 
- Using the SPDX identifier for the license. 
- Added a list of the most important dependencies before the build. 
- Code highlighted `mvn clean install` because it's an important one. 
- Added further links to docs. 

I hope that's fine with you. Or do you wish any adjustments?

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,31 @@ Mustangproject
 
 Source code repository for the [Mustang project](http://www.mustangproject.org/) open source java PDF invoice metadata library in ZUGFeRD format.
 
-License
+Build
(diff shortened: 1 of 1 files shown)


jstaerk:
actually thats fine and much more concise than I would have had the time for :-) Thanks!

PR #367, opened 2024-04-12
handle read from empty file

hduelme:
I added a check if the file to read from is empty if yes a error is reported. In addition I use `Arrays.equals` to check if the first 4 bytes match a PDF-file.


jstaerk:
thanks!

PR #382, opened 2024-05-15
Added getter for validation context to get the ValidationIssues Vector for further processing in another application

Sarenor:


jstaerk:
thank you, after merging I moved getValidationContext into the Validator class where both XMLValidator and ZUGFeRDValidator inherit

Sarenor:
Makes absolute sense, thanks!

PR #383, opened 2024-05-17
#337_ Be-able-to-validate-XRechnung-UBL-files

5now:
Suggestion for a new composition of the pom.xml


phax:
The source issues was already resolved, so I think this PR may be closed

Issue #387, opened 2024-05-29
PDF generation from XML requires FreeSans.ttf


Heavenfighter:
The action pdf of the CLI which transforms XML to PDF throws an error, that the FreeSans.ttf is missing. It is not sufficient to install that font, it must explicitly be present in the calling directory.

Tested on Windows 10 system.

`
2024-05-29 08:17:48.379 [main] ERROR ROOT - Failed to read font file FreeSans.ttf C:\Users\user\Desktop\FreeSans.ttf (Das System kann die angegebene Datei nicht finden)
java.lang.RuntimeException: Failed to read font file FreeSans.ttf C:\Users\user\Desktop\FreeSans.ttf (Das System kann die angegebene Datei nicht finden)
	at org.apache.fop.fonts.LazyFont.load(LazyFont.java:134)
	at org.apache.fop.fonts.LazyFont.getAscender(LazyFont.java:237)
	at org.apache.fop.fonts.Font.getAscender(Font.java:120)
	at org.apache.fop.layoutmgr.BlockLayoutManager.initialize(BlockLayoutManager.java:82)
	at org.apache.fop.layoutmgr.AbstractLayoutManager.getChildLM(AbstractLayoutManager.java:118)
	at org.apache.fop.layoutmgr.FlowLayoutManager.getNextKnuthElements(FlowLayoutManager.java:108)
	at org.apache.fop.layoutmgr.FlowLayoutManager.getNextKnuthElements(FlowLayoutManager.java:69)
	at org.apache.fop.layoutmgr.PageBreaker.getNextKnuthElements(PageBreaker.java:252)
	at org.apache.fop.layoutmgr.AbstractBreaker.getNextBlockList(AbstractBreaker.java:675)
	at org.apache.fop.layoutmgr.PageBreaker.getNextBlockList(PageBreaker.java:179)
	at org.apache.fop.layoutmgr.PageBreaker.getNextBlockList(PageBreaker.java:159)
	at org.apache.fop.layoutmgr.AbstractBreaker.doLayout(AbstractBreaker.java:385)
	at org.apache.fop.layoutmgr.PageBreaker.doLayout(PageBreaker.java:113)
	at org.apache.fop.layoutmgr.PageSequenceLayoutManager.activateLayout(PageSequenceLayoutManager.java:143)
	at org.apache.fop.area.AreaTreeHandler.endPageSequence(AreaTreeHandler.java:267)
[...]

PR #388, opened 2024-05-30
#387 allow embedd fonts from jar-file for pdf creation

Heavenfighter:
fonts can now added in fop configuration via classpath notation. 
Also added FreeSans.ttf


jstaerk:
Incredible. I could swear I tried that. Verapdf reports two glitches in the PDF/A file but I consider to just upstream that to the Fop guys 
[Validation Report.zip](https://github.com/user-attachments/files/15509989/Validation.Report.zip)

Heavenfighter:
Maybe one could move the font in the font directory, but i couldn't figure out how that is done.

PR #393, opened 2024-06-18
Support for Jakarta and PDFBox 3.x

phax:
@jstaerk  please find attached a PR that uses Jakarta instead of javax and also updates to PDFBox to 3.x (required because PDFBox 2.x uses javax.).

All unit tests pass in the IDE.

One thing I needed to do, is to disable compression in the created PDFs so that the metadata tests work. I propose to remove this test and enable PDF compression instead....

I also took the liberty to change the version to 3.0.0-SNAPSHOT because it's a lot of breaking changes...

Edit: Updating the VeraPDF dependency from `validation-model` to `validation-model-jakarta` resolved the last issue I found (see https://github.com/veraPDF/veraPDF-library/issues/1314)

### DIFF
--- a/Mustang-CLI/pom.xml
+++ b/Mustang-CLI/pom.xml
@@ -30,24 +30,24 @@
       <dependency><!-- apache commons cli to parse command line -->
          <groupId>commons-cli</groupId>
          <artifactId>commons-cli</artifactId>
-         <version>1.5.0</version>
+         <version>1.8.0</version>
       </dependency>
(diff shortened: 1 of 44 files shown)


jstaerk:
Very much appreciated, thanks, I believe I read you had been in vacation already, I still am away this week. But it wont be forgotten.

phax:
@jstaerk  enjoy your vacation :)

PR #394, opened 2024-06-20
Re-Added getValidationContext to ZUGFeRD Validator

Sarenor:
I'm not sure if it was an intended change to remove the Getter I added but with the ZUGFeRD validator no longer implementing the Validator Interface it'd still be nice to be able to get the context to easily get the ResultItems during processing.


jstaerk:
I believe you had put the same function in pdfvalidator and xmlvalidator and I moved it to the validator class, from which both pdfvalidator and xmlvalidator inherit the function?

Sarenor:
https://github.com/ZUGFeRD/mustangproject/pull/382/files
To the Zugferd Validator, actually, since it's the file I would be calling instead of calling the PDFValidator directly.

PR #406, opened 2024-07-09
Misc optimizations

phax:
* Require Java 11 as the minimum - otherwise Jakarta and other stuff won't work properly
* Allow a validation of Zugferd documents without requiring a file - an `InputStream` is sufficient. This also reduces the amount of IO operations needed to perform the activities
* Make sure the Zugferd validation APIs work in the same way
* Make some general code improvements (constant instead of `UTF-8`; remove unnecessary code; remove unnecessary imports)
* Remove an additional dependency to the `BigFileSearcher` as the data was already present as a `byte[]` so the `ByteArraySearcher` is more efficient
* Removed unused variables and unused private members
* Removed unused casts
* Removed unnecessary duplicate "implements"


jstaerk:
cosmetical question: debCN, debCurrentChild sounds like "deb"ug?
setOwnTaxID etc will continue to work? You might have noticed that I would very much like to publish as (backward compatible) Mustang 2.12 not yet Mustang 3.
I would like to close this branch and only accept the new ZUGFeRD version when release, if there are further bigger refactorings or changes maybe we could start a 2.13 branch.

phax:
@jstaerk  well, I just reused the `deb` variables, as a performance improvement - they all existed beforehand. Do you want me to rename them to not contain `deb`?
The `setOwnTaxID` methods etc. are of course still working. They are just marked as deprecated, so I replaced them with functionally-equal non-deprecated methods :)
Hope that solves your questions.

PR #407, opened 2024-07-10
Replace java.util Logging with SLF4J

phax:
* Replace all occurrances of java.util.logging stuff with SLF4J
* Added initial messages instead of "null" loggings
* Remove double logging, where it was evident that it does not add value


phax:
Part of the [2.12.0](https://github.com/ZUGFeRD/mustangproject/releases/tag/core-2.12.0) release

PR #410, opened 2024-07-11
Return all BankDetails from parsed CII xml. Closes #408.

### LINKED ISSUE #408: Invoice extracted from CII XML returns only one IBAN (BT-84) although there are more
Given a CII XML file from [itplr-kosit/xrechnung-testsuite](https://github.com/itplr-kosit/xrechnung-testsuite/blob/master/src/test/business-cases/standard/01.05a-INVOICE_uncefact.xml) when extracting an Invoice with `org.mustangproject.ZUGFeRD.ZUGFeRDInvoiceImporter#extractInvoice` and accessing `org.mustangproject.TradeParty#getBankDetails` through `org.mustangproject.Invoice#getSender` then only one `org.mustangproject.BankDetails` object is returned.

Because the CII XML actually contains `PayeePartyCreditorFinancialAccount` twice (each within its own `SpecifiedTradeSettlementPaymentMeans`), I would have expected `invoice.getSender().getBankDetails()` to return a list with two `org.mustangproject.BankDetails` objects.

langfr:


jstaerk:
cool thanks a lot!

PR #432, opened 2024-07-30
Enhance Charges/Allowances with reasonCode.

langfr:
Added field reasonCode to IZUGFeRDAllowanceCharge, used for Charges and Allowances.
ZUGFeRDInvoiceImporter: read the value during import.
ZUGFeRD2PullProvider:
Export Allowances and Charges one by one, not aggregated by taxPercent, but only for profile XRechnung for now.
Don't know if this is also allowed for other profiles also.
Added reasonCode also on item level.
Removed some obsolete spaces in xml-Fragments.


jstaerk:
thanks!

PR #446, opened 2024-08-15
Re-Formatted and re-organized POMs

hypfvieh:
This PR includes a refactoring of the project POMs.

- Formatted all POMs using Maven sortpom Plugin
- Removed duplicated configuration (will be inherited from the parent POM)
- Moved all plugins to plugin management with a base (default) configuration
- Use plugins in modules by referencing them, added additional configuration if needed
- Updated plugin versions
- Add test-jar to created jars on build (I wonder why Maven Central did not complain about that)
- Moved common dependencies to dependency-management
- Added properties for all used dependency versions
  - this allows defining the versions only once and at a well defined position

I build everything locally and also did `release:prepare` and `release:perform`. 
The created artifacts looks good to me. POM seems to be complete (not missing dependencies).  
I don't know how exactly your build-chain works, but usually the maven-deploy plugin should deploy to Maven Central, while it will only deploy to some local directories with the current configuration (like before).

One more thing I've noticed: The surefire plugin is configured to run the tests in a specified order. That is really awkward. Tests should never rely on each other, every test class should be able to perform the required actions without any knowledge of other tests. E.g. this setup will not work when running tests in an IDE, because some resources would be missing when another test was not executed explicitly before.

I did not change that behavior even though I was thinking about it, but that would at least be stuff for another PR and also would require refactoring in test setup.

If you need further assistance with this, let me know.

### DIFF
--- a/Mustang-CLI/pom.xml
+++ b/Mustang-CLI/pom.xml
@@ -1,215 +1,87 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
     <parent>
         <groupId>org.mustangproject</groupId>
         <artifactId>core</artifactId>
         <version>2.13.0-SNAPSHOT</version>
     </parent>
-    <modelVersion>4.0.0</modelVersion>
+
     <artifactId>Mustang-CLI</artifactId>
-    <name>e-invoices commandline tool, allowing to create(embed), split and validate Factur-X/ZUGFeRD files. Validation
-        should also work for XRechnung/CII.
-    </name>
     <packaging>jar</packaging>
+    <name>E-Invoice commandline tool</name>
+    
+    <description>
+    	E-invoices commandline tool, allowing to create(embed), split and validate Factur-X/ZUGFeRD files. Validation
+        should also work for XRechnung/CII.
+    </description>
+    
     <properties>
-        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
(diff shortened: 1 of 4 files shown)


jstaerk:
Hi,
the test order is indeed required, AFAIK that was a hack we implemented because the files tested by the validator first need to be generated using the library.

I didnt merg it right away because not only of the release-sign-artifacts profile but also the distributionManagement section seemed to have been removed?

Anyway I hope I can have a closer look tomorrow.

kind regards
Jochen

hypfvieh:
Could you please explain which part of the release cycle does not work?
How does your release cycle looks like?

One thing I came across while doing this changes was the configuration of the maven-deploy plugin which always "uploaded" the artifacts to a local folder instead of uploading it to Maven Central (see line 157 in "core" pom.xml of my PR).
As I don't know your way of releasing, I left that configuration like it was (personally I would remove it).

In my projects (e.g. [dbus-java)](https://github.com/hypfvieh/dbus-java), I use `mvn release:prepare` and `mvn release:perform` to do the release stuff. 
My release cycle looks like this:
`mvn release:prepare`
`git push --follow-tags`
`mvn release:perform`

The reason why I do the `git push` manually is I can avoid adding tags and failed-release-checkins upstream when the build during `prepare` step failed (e.g. unit test failure or other issues like checkstyle, PMD etc). 
I simply delete the tag and revert the changes of `mvn release:prepare` by resetting the branch back to upstream master status.

When everything works, so I could `prepare`, `push` and `release`, I will take a look at Maven Central to confirm my release.
This is a matter of taste, you can also configure the plugin to do this for you automatically.

jstaerk:
Hello,
>Could you please explain which part of the release cycle does not work?
It did not upload anything at all to maven central.

>How does your release cycle looks like?
I run `mvn release:prepare` and `mvn release:perform` which automatically builds, signs, uploads and publishes on maven central, afterwards I upload the created mustang-cli.jar (+the lib and the validator) to mustangproject.org, update the homepage, create a post in german and english based on the history.md, publish the new release on freshnode and upload mustang-cli.jar to a new release on github.

kind regards
Jochen

Issue #448, opened 2024-08-19
class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA cannot be cast to class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3


msccip:
Mustang-CLI-2.12.0.jar: If you try to add an additional PDF file as an attachment using the combine action, the following error occurs:

class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA cannot be cast to class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3 (org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA and org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3 are in unnamed module of loader 'app')

the following commandline is used:

java -jar Mustang-CLI-2.12.0.jar --action combine --source x.pdf --source-xml y.xml --out rechnung.pdf --format zf --version 2 --profile E --attachments "z.PDF"

A stacktrace is not displayed. With version 2.10 it works without problems

msccip:
With version 2.13 the exception no longer appears. But attachments are not added at all. They are simply ignored.

msccip:
fixed in 2.14

Issue #389, opened 2024-06-01
ClassCastException: ZUGFeRDExporterFromA3 cannot be cast to ZUGFeRDExporterFromPDFA


langfr:
When using
            `IZUGFeRDExporter ze = new ZUGFeRDExporterFromPDFA();`
then
`            ze.setProfile( Profiles.getByName( "extended" ) );`
leads to a class cast exception:
```Caused by: java.lang.ClassCastException: org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3 cannot be cast to org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA
	at org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA.setProfile(ZUGFeRDExporterFromPDFA.java:151)
	at org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA.setProfile(ZUGFeRDExporterFromPDFA.java:40)
	at biz.mbisoftware.fn.ejb.session.sales.InvoiceExport.buildInvoiceZugferdPdf(InvoiceExport.java:523)
	at biz.mbisoftware.fn.ejb.session.sales.InvoiceExport.processInvoice(InvoiceExport.java:403)
```

msccip:
I get the opposite message with version 2.12.0: ...FromPDFA cannot be cast to class ...FromA3

class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA cannot be cast to class org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3 (org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromPDFA and org.mustangproject.ZUGFeRD.ZUGFeRDExporterFromA3 are in unnamed module of loader 'app')

langfr:
Please create a new issue.
Where does this Exception exactly happen ? Main.java:726 ? The complete stacktrace would be helpful.
Share the command line you're using, so that somebody is able to reproduce this issue. java -jar Mustang-CLI-2.12.0.jar .... ?

PR #451, opened 2024-08-20
Fix ClassCastException in CLI (Main.java).

langfr:
Related to #448 .

Moved adding of Attachments after loading the pdf-file. Only at this point it's clear, of which type the source pdf is.
Check if exporter class is really of type ZUGFeRDExporterFromA3, else (ZUGFeRDExporterFromA1) print an error message.
At least with a source pdf of PDF/A-3A it works.
Class ZUGFeRDExporterFromA1 does not support Attachments, because in PDF/A-1 (ISO 19005-01) version 1.4 (2005) Attachments are NOT allowed.

```
java -jar Mustang-CLI-2.13.0-SNAPSHOT.jar --action combine --source Rechnung.pdf --source-xml CII-XRechnung-4711-0815-xml.xml --out ZF_Rechnung.pdf --format zf --version 2 --profile E  --attachments "CII-XRechnung-4711-0815-xml.pdf"

SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
Source PDF set to Rechnung.pdf
ZUGFeRD XML set to CII-XRechnung-4711-0815-xml.xml
Output PDF set to ZF_Rechnung.pdf
Format set to zf
Version set to 2
Profile set to E
Written to ZF_Rechnung.pdf
```

### DIFF
--- a/Mustang-CLI/src/main/java/org/mustangproject/commandline/Main.java
+++ b/Mustang-CLI/src/main/java/org/mustangproject/commandline/Main.java
@@ -722,9 +722,6 @@ public class Main {
 					((ZUGFeRDExporterFromPDFA) ze).ignorePDFAErrors();
 				}
 			}
-			for (FileAttachment attachment : attachments) {
-				((ZUGFeRDExporterFromA3) ze).attachFile(attachment.getFilename(), attachment.getData(), attachment.getMimetype(), attachment.getRelation());
-			}
 
 			ze.load(pdfName);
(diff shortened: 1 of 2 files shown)


jstaerk:
Hi, wasn't that fixed in #389 already?
FYI: no matter if the *input* was A1 or A3, the *output* of Mustang will always be A3, so the output will always support embedded files.

Issue #458, opened 2024-08-30
Its not possible to add multiple PaymentTerms when using the extended profile


Frgetg:
The Invoice class only allows setting a single PaymentTerm at most, whcih is fine for all profiles apart from extended:

![image](https://github.com/user-attachments/assets/466efbe5-6bce-4e1d-b1b0-22d3663f3243)

I need to add several payment terms to add multiple cash discount conditions.

Frgetg:
Pull Request: https://github.com/ZUGFeRD/mustangproject/pull/459

damoasda:
It is working now, use `Invoice.setExtendedPaymentTerms()`. This issue can be closed.

PR #459, opened 2024-08-30
Isue #458 Allow Multiple payment details for invoice

Frgetg:


jstaerk:
@Frgetg  Good idea my issue is that you are changing public methods which would require a new major version of the software, can't you just add something like addPaymentTerms instead of changing setPaymentTerms to an array?

Frgetg:
I can change the setter, yes, however, the getter will always have to return the whole array, or the XML serialization wont access all PaymentTerms.

And both a getPaymentTerms which always returns only the first one, and getAllPaymentTerms which returns everything feels rather hacky and error prone.

marco-lennartz:
@jstaerk how would you like to deal this? In the extended profile are 0...n SpecifiedTradePaymentTerms allowed. So the payment terms must be an array or a list. As @Frgetg stated handling a getter with only one PaymentTerm would be very error prone.

jstaerk:
>a getter with only one PaymentTerm would be very error prone
if that getter already exists as a public method the maximum you can 

* either wait for the next major version, maybe in some years OR 
* **add** an alternative
* continue behaving like previously with what **is** implemented
* the maximum you could do is depreate the old getter

Frgetg:
Hello @jstaerk 

I adapted the patchset according to your guidelines

schliepRon:
@jstaerk whats the status of the PR? Will the feature be officially available soon?

jstaerk:
apologies for the delay

PR #470, opened 2024-09-10
Add support for BuyerOrderReferencedDocument on line level

### LINKED ISSUE #466: BuyerOrderReferencedDocument / IssuerAssignedID
In #247 wurde die Möglichkeit geschaffen, die Zeilennummer der Bestellung anzugeben. Das reicht für uns leider nicht, da wir häufig mehrere Bestellungen (nicht mehrere Lieferungen zu einer Bestellung) auf einer Rechnung haben. Wir lösen das im Augenblick mit einem Freitext (in `getNotes`), das ist allerdings eher eine Notlösung.

Prinzipiell könnte man das natürlich genauso lösen wie in #247, das ist aber etwas unschön, weil die API damit zukünftig sehr unübersichtlich wird (wenn jemand auch `FormattedIssueDateTime` oder etwas aus dem Bereich `QuotationReferencedDocument` nutzen möchte). IMO wäre die saubere Implementierung hier in `IDatedReferencedDocument` mit `getIssuerAssignedId`, `getLineID` und `getFormattedDate`. Technisch ist das nicht wahnsinnig kompliziert, würde aber entweder zu doppelter Funktionalität führen (wenn `getBuyerOrderReferencedDocumentLineID` erhalten bleibt) oder wäre nicht backward-compatible.

Edit: added translation

#247 added the option to use the order line id. Unfortunately that is not enough for us since we often include different orders (not just several shipments related to one order) in one invoice. We currently use a free text (in `getNotes`) but this is only a workaround.

In principle this could be solved similar to #247, but that's not very nice, it clutters the API (especially if future additions for `FormattedIssueDateTime` or something from `QuotationReferencedDocument` are requested). IMO the cleanest implementation would be a `IDatedReferencedDocument` with `getIssuerAssignedId`, `getLineID` and `getFormattedDate`. Technically that's not challenging, but it either duplicates existing functionality (if `getBuyerOrderReferencedDocumentLineID` stays in API) or is nor backward-compatible.

J-N-K:
Closes #466

### DIFF
--- a/library/src/main/java/org/mustangproject/Item.java
+++ b/library/src/main/java/org/mustangproject/Item.java
@@ -22,7 +22,8 @@ public class Item implements IZUGFeRDExportableItem {
 	protected BigDecimal basisQuantity = BigDecimal.ONE;
 	protected Date detailedDeliveryPeriodFrom = null, detailedDeliveryPeriodTo = null;
 	protected String id;
-	protected String referencedLineID = null;
+	protected String buyerOrderReferencedDocumentLineID = null;
+	protected String buyerOrderReferencedDocumentID = null;
 	protected Product product;
 	protected ArrayList<String> notes = null;
 	protected ArrayList<ReferencedDocument> referencedDocuments = null;
@@ -61,7 +62,8 @@ public class Item implements IZUGFeRDExportableItem {
 		String vatPercent = null;
 		String lineTotal = "0";
 		String unitCode = "0";
-		String referencedLineID = null;
+		String buyerOrderReferencedDocumentLineID = null;
+		String buyerOrderReferencedDocumentID = null;
 
 		ArrayList<ReferencedDocument> rdocs = null;
 		ArrayList<ReferencedDocument> addRefs = null;
@@ -150,7 +152,9 @@ public class Item implements IZUGFeRDExportableItem {
 						for (int docIndex = 0; docIndex < docChilds.getLength(); docIndex++) {
 							String localName = docChilds.item(docIndex).getLocalName();
 							if ((localName != null) && (localName.equals("LineID"))) {
-								referencedLineID = docChilds.item(docIndex).getTextContent();
+								buyerOrderReferencedDocumentLineID = docChilds.item(docIndex).getTextContent();
+							} else if ((localName != null) && (localName.equals("IssuerAssignedID"))) {
+								buyerOrderReferencedDocumentID = docChilds.item(docIndex).getTextContent();
 							}
 						}
 					}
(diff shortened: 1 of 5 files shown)


library/src/main/java/org/mustangproject/Item.java
```diff
@@ -310,22 +314,34 @@ public Item(NodeList itemChilds, boolean recalcPrice) {
...
+	public Item addBuyerOrderReferencedDocumentLineID(String s) {
+		buyerOrderReferencedDocumentLineID = s;
+		return this;
+	}
+
+	@Deprecated(since = "2.14.0")
 	public Item addReferencedLineID(String s) {
-		referencedLineID = s;
+		return addBuyerOrderReferencedDocumentLineID(s);
+	}
```
J-N-K, line 330: I renamed this method to align it with the getter (which is part of the interface) and also be more consistent. I have kept and deprecated the old method for backward compatibility.

library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
```diff
@@ -448,10 +448,17 @@ public void generateXML(IExportableTransaction trans) {
 							"</ram:AdditionalReferencedDocument>";
 					}
 				}
-				if (currentItem.getBuyerOrderReferencedDocumentLineID() != null) {
-					xml += "<ram:BuyerOrderReferencedDocument> "
-						+ "<ram:LineID>" + XMLTools.encodeXML(currentItem.getBuyerOrderReferencedDocumentLineID()) + "</ram:LineID>"
-						+ "</ram:BuyerOrderReferencedDocument>";
+				if ((currentItem.getBuyerOrderReferencedDocumentLineID() != null) || (currentItem.getBuyerOrderReferencedDocumentID() != null)) {
+					xml += "<ram:BuyerOrderReferencedDocument> ";
```
J-N-K, line 452: As a general comment: wouldn't it make sense to use `StringBuilder` in this class instead of creating a new `String` object with each step?

J-N-K (commented):
`\library\src\main\java\org\mustangproject\ZUGFeRD\ZUGFeRDInvoiceImporter.java` doesn't require modification because this is already handled by `Item`.

library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java
jstaerk: yes stringbuilder would make sense

jstaerk:
apologies for the late reaction, do you have time to resolve the conflicts?

PR #472, opened 2024-09-10
Fix logging implementation missing in CLI

### LINKED ISSUE #434: mustang-cli 2.12.0 warns about missing SLF4J binding; no log output in file anymore
Hi,
since the commit https://github.com/ZUGFeRD/mustangproject/pull/407/commits/0f5ff080d8f91ac527530a1a18da3ed4757968fa the releases do not contain the jar file for SLF4J and it is not findable in pom.xml. The compiled jar cannot be run on a console without warning:

```
2024-08-05 20:47:43,108 | ERROR | stderr: SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

since that change, log output is not written to default dir ./log/ZUV-*.log

i am not a java developer. i am an enduser using mustang to create and validate hybrid pdf invoices.

J-N-K:
Fixes #434 

No logging implementation was included in the shaded jar resulting in the mentioned error message and missing log outputs.

### DIFF
--- a/Mustang-CLI/pom.xml
+++ b/Mustang-CLI/pom.xml
@@ -35,6 +35,12 @@
             <version>1.8.0</version>
         </dependency>
 
+		<dependency>
+			<groupId>org.slf4j</groupId>
+			<artifactId>slf4j-simple</artifactId>
+			<version>2.0.12</version>
+		</dependency>
+
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>


jstaerk:
@phax did we remove this on purpose?

J-N-K:
It was never added. When switching from `java.util`-logging to slf4j you need to include the api and one of several implementations. slf4j-simple is quite common for standalone-cli-applications (it outputs to `System.err`), but you could also choose something like log4j or logback. That might require some configuration but also allows logging to a file.

phax:
@jstaerk makes sense to me

PR #477, opened 2024-09-13
Fix prepaid amount in importer

### LINKED ISSUE #475: Example invoice coud not be parsed
While implementing in Jes, I used the [example invoices from OpenXRechnungToolbox](https://github.com/jcthiele/OpenXRechnungToolbox/tree/master/exampleInvoices/XRechnung_v2.2.0) for my tests. Most of them work well, but I got a ParseException for 04.01a-INVOICE_ubl.xml.

Could you check what's wrong?

J-N-K:
Fixes #475

### DIFF
--- a/library/src/main/java/org/mustangproject/ZUGFeRD/TransactionCalculator.java
+++ b/library/src/main/java/org/mustangproject/ZUGFeRD/TransactionCalculator.java
@@ -39,8 +39,8 @@ public class TransactionCalculator implements IAbsoluteValueProvider {
 	}
 
 	/***
-	 * the invoice total with VAT, corrected by prepaid amount, allowances and
-	 * charges
+	 * the invoice total with VAT, allowances and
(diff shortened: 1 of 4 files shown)


library/src/main/java/org/mustangproject/ZUGFeRD/TransactionCalculator.java
```diff
@@ -39,8 +39,8 @@ protected BigDecimal getTotalPrepaid() {
 	}
 
 	/***
-	 * the invoice total with VAT, corrected by prepaid amount, allowances and
-	 * charges
+	 * the invoice total with VAT, allowances and
+	 * charges, WITHOUT considering prepaid amount
```
J-N-K, line 43: This was wrong JavaDoc. The prepaid amount is not considered when calculating the grand total (and this is the expected behavior by other methods).

PR #488, opened 2024-09-20
Fix newly introduced validation failures.

langfr:
<ram:PayeeTradeParty> at wrong location (???).
Parameters for LegalOrganisation need to be switched (#471).


J-N-K:
LGTM.

With regard to `LegalOrganisation`: #457 introduced the new test and was merged after I created #471. That's why it wasn't fixed.

langfr:
Fixed by commit https://github.com/ZUGFeRD/mustangproject/commit/12315cf7e715403932d37e8598c8da9cdcbc41e9.

Issue #494, opened 2024-09-27
Quantity Decimal Places Limitation Conflicts with XRechnung Compliance (Needs Immediate Fix)


fktrdui:
According to the XRechnung specification, as highlighted in the attached screenshot (section 8.8 "Quantity"), the quantity values should not have a limitation on decimal places. The specification explicitly mentions that the quantity field should allow a floating-point number without limitation on decimal places ("ohne Limitierung der Nachkommastellen").
![quantityFormat](https://github.com/user-attachments/assets/f9591d2f-2b4b-45ac-8c57-2b0e02296013)

However, the Mustang project appears to limit the number of decimal places for quantities in certain methods. 
Specifically, in the ZUGFeRD2PullProvider class, the quantityFormat method restricts quantities to 4 decimal places:
```
protected String quantityFormat(BigDecimal value) {
    return XMLTools.nDigitFormat(value, 4);
}
```
This method is called by:
```
+ "<ram:BasisQuantity unitCode=\"" + XMLTools.encodeXML(currentItem.getProduct().getUnit())
+ "\">" + quantityFormat(currentItem.getBasisQuantity()) + "</ram:BasisQuantity>"

```
Additionally, in the IZUGFeRDExportableItem interface, the getBasisQuantity() method also uses BigDecimal.ONE.setScale(4), which imposes a restriction of 4 decimal places for the quantity.
```
default BigDecimal getBasisQuantity() {
    return BigDecimal.ONE.setScale(4);
}
```

**Request for Immediate Action**
This issue requires urgent attention as it directly impacts XRechnung compliance. Non-compliant e-invoices could lead to legal and financial risks, such as invoice rejections.

**Proposed Fix**: 
Modify the following methods to allow arbitrary precision for quantity values:
quantityFormat in the ZUGFeRD2PullProvider class.
getBasisQuantity in the IZUGFeRDExportableItem interface.

fktrdui:
Thank you for your immediate update.
Due to changes there are some test issues in the project:

Failed tests:   testDXExport(org.mustangproject.ZUGFeRD.DXTest): expected:<400.0000> but was:<400.000000000000000000>
  testOXEdgeExport(org.mustangproject.ZUGFeRD.OXTest): expected:<1.0000> but was:<1.000000000000000000>
  testOXExport(org.mustangproject.ZUGFeRD.OXTest): expected:<400.0000> but was:<400.000000000000000000>
[...]

PR #497, opened 2024-09-30
Fix validation errors introduced by fix for #494.

langfr:


jstaerk:
thanks but I'm thinking if we cant display 4 digits if the other ones are 0 anyway

jstaerk:
as mentioned, thanks, biut I introduced a nDigitNumberFormatDecimalRange in which by default now at least 4 digits are shown after the comma, and more if needed

PR #508, opened 2024-10-07
Update documentation for `IExportableTransaction.getReferenceNumber()`

### LINKED ISSUE #504: Misleading documentation for `IExportableTransaction.getReferenceNumber`
https://github.com/ZUGFeRD/mustangproject/blob/6dcf7f50d9b64e3abce056e9db11cfb348896c8b/library/src/main/java/org/mustangproject/ZUGFeRD/IExportableTransaction.java#L324-L330

This is used for `BuyerReference` BT-10, and according to at least the German norm that seems like a strange documentation:

> eine vom Käufer zugewiesene und für internes Routing benutzte Kennung

> an identifier assigned by the buyer and used for internal routing

WorkProg:
fixes #504


jstaerk:
Sorry but buyerReference is really more "your reference", mostly in terms of "care of", in XRechnung used for the recipient ID (leitweg id). When something refers to a previous order, like a cancellation or a corrected invoice it should rather use a InvoiceReferencedDocument, like correctedInvoices do in mustang.

WorkProg:
That is what I meant, maybe you can word it better, but the current documentation is

>  get reference document number typically used for Invoice Corrections Will be added as IncludedNote in comfort profile

Which is as you seem to agree wrong.

WorkProg:
Therefor, I wanted to change it to:

> get BuyerReference (BT-10) an identifier assigned by the buyer and used for internal routing. Used for the Leitweg-ID.

PR #526, opened 2024-10-21
Trim values of whitespace before parsing

### LINKED ISSUE #525: Extra whitespace causing format exceptions
Hey everyone,

We have a case where whitespace is causing `BigDecimal` conversion errors - see [here](https://github.com/ZUGFeRD/mustangproject/blob/52ddff4f7ac0bf9ee7f63854499c73ee503de767/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDInvoiceImporter.java#L328)

Obviously, the creator of the e-invoice shouldn't be adding extra whitespace around values, but alas, here we are. Other tools seem to handle this gracefully. I would like to submit a PR with whitespace trimmed but figured I'd check in here first.

lolletsoc:
A format has gone over this file - if you're not happy with it, I'll revert the changes

Fixes #525


jstaerk:
thank you

PR #588, opened 2024-11-26
Extend invoice creation with json deserialization

phermsdorf:
This pull request add JSON serialization and deserialization support for the following members of the Item class:
* BuyerOrderReferencedDocumentLineID
* Allowances
* Charges

I extended the DeSerializationTest to verify the changes and show the usage.

### DIFF
--- a/library/src/main/java/org/mustangproject/Item.java
+++ b/library/src/main/java/org/mustangproject/Item.java
@@ -1,7 +1,11 @@
 package org.mustangproject;
 
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
-import com.fasterxml.jackson.annotation.JsonInclude;
+import java.math.BigDecimal;
+import java.math.RoundingMode;
+import java.util.ArrayList;
+import java.util.Date;
+import java.util.List;
+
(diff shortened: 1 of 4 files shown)


jstaerk:
not sure about BuyerOrderReferencedDocumentLineID but I already did charges/allowances (sorry), could you please check and resolve? thanks and apologies.

phermsdorf:
Hi @jstaerk ,

i have seen you did charges and allowances for the **invoice**, my change/extension is for the **item**.

I could revert the name change of addReferencedLineID to setBuyerOrderReferencedDocumentLineID and just add the set method if that would be better.

Should i update my pull request to the current code with that changes in place?

Thansk, Bye Peter

jstaerk:
No matter what you do we need to at least stay compatible to the old public methods (and properties) or this has to wait for the next major version (maybe in a couple of years). e.g. public Item addReferencedLineID vs
setBuyerOrderReferencedDocumentLineID(String s) but also regards charges vs ItemCharges: actually I never tested that you have a chance that it never worked so far in which case you could of course decide for a name

phermsdorf:
this now seems to be completed with #761

PR #590, opened 2024-11-27
allow setting payment means and payment means description, fixes #516

### LINKED ISSUE #516: SpecifiedTradeSettlementPaymentMeans hardcoded
Why are these hardcoded in IZUGFeRDTradeSettlementPayment.java? (I know I can overwrite it if I want to)

`		String xml = "<ram:SpecifiedTradeSettlementPaymentMeans>"
				+ "<ram:TypeCode>58</ram:TypeCode>"
				+ "<ram:Information>SEPA credit transfer</ram:Information>"`

ThomasChr:
I tried to fix #516 and allow to set payment means code and payment means description.
I'm not a Java Guy, so be gentle when reviewing :-)

### DIFF
--- a/library/src/main/java/org/mustangproject/DirectDebit.java
+++ b/library/src/main/java/org/mustangproject/DirectDebit.java
@@ -21,6 +21,16 @@ public class DirectDebit implements IZUGFeRDTradeSettlementDebit {
 	protected String mandate;
 
 	/**
+	 * payment means code (BT-81 / UNTDID 4461)
+	 */
+	protected String paymentMeansCode = "59";
+
+	/**
+	 * Payment means description (BT-82)
+	 */
+	protected String paymentMeansInformation = "SEPA direct debit";
+
+	/**
 	 * bean constructor
 	 */
 	public DirectDebit() {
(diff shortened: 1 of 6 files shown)


ThomasChr:
I also added reading of the payment means code and payment means information

jstaerk:
Is paymentMeansInformation/paymentMeansCode part of direct debit, of any payment means, or of the invoice? Would it npot be duplicated in the invoice when you read and in the direct debit when you write?

ThomasChr:
@jstaerk 
Thanks for taking a look!
Yeah, having that in DirectDebit and Invoice is a duplicate.
I killed it from Invoice, I think it's best at DirectDebit.

What do you think?

ThomasChr:
If I read the standard correctly payment means should be on invoice level. Should I change the code to be on Invoice level or is the direct debit approach good for now?

jstaerk:
good idea+implementation but why was the sepa debit code changed from 59 to 58?

ThomasChr:
> good idea+implementation but why was the sepa debit code changed from 59 to 58?

You absolutely right, sorry for that and thanks for taking a look!
Fixed the 58/59 and rebased the branch!

Paeprock:
Wouldn't this rather be needed in BankDetails instead of DriectDebit? Since DirectDebit should always have the direct debit payment means?
Because my invoices are not paid by direct debit but rather by a manual bank transfer.

ThomasChr:
@Paeprock 
I used the methods and hat the same idea. I tend to think the same.
@jstaerk Should I move the methods to BankDetails, what do you think?

PR #591, opened 2024-11-27
Import IncludedNotes on item extraction

InnuceEAN:
similar to #554 but on item level


jstaerk:
@BhardwajB could you have a look at UBL?

BhardwajB:
> @BhardwajB could you have a look at UBL?

sure.

PR #595, opened 2024-11-29
Treat all fatal XR schematron rules as errors, not as warnings.

mkilian:
In org.mustangproject.validator.XMLValidator.validateSchematron(), the severity has to be determined for every node of the validation result.  Otherwise, all schematron assertions tagged as "fatal" will be treated as warnings after the first warning.


mkilian:
Some more details:

In commit aeb491873dbc7486cfc21d7d0609561e74942f55, the severity
is only properly initialized (to ESeverity.error), which, after the
first failed assertion tagged as "warning" found in the XR schematron
validation result, causes all further failed assertions to be treated
as warnings, regardless wether they are tagged as "warning" or
"fatal."

For example, validator/src/test/resources/invalidXRV30.xml is missing
the buyer reference (BT-10), but the related assertion [BR-DE-15]
is counted as a warning only.

This change moves the determination of the severity to apply into
the loop over the failed assertions of the schematron validation
and also changes the corresponding test case accordingly.

PR #597, opened 2024-12-03
Reuse toPDF method to work without any dependencies to the file system

taskinkemal:


jstaerk:
looks good can you have a look to resolve the conflicts or shall I?

Issue #605, opened 2024-12-06
Can set the document name in an invoice, but it will not be written into the XML.


haba-beton:
Invoice invoice = new Invoice();
invoice.setDocumentName("RECHNUNG");

XML generation inside of ZUGFeRD2PullProvider: Line 385

		xml +=
			"<ram:GuidelineSpecifiedDocumentContextParameter>"
				+ "<ram:ID>" + getProfile().getID() + "</ram:ID>"
				+ "</ram:GuidelineSpecifiedDocumentContextParameter>"
				+ "</rsm:ExchangedDocumentContext>"
				+ "<rsm:ExchangedDocument>"
				+ "<ram:ID>" + XMLTools.encodeXML(trans.getNumber()) + "</ram:ID>"
				// + "<ram:Name>RECHNUNG</ram:Name>"
				// + "<ram:TypeCode>380</ram:TypeCode>"
				+ "<ram:TypeCode>" + typecode + "</ram:TypeCode>"
				+ "<ram:IssueDateTime>"
				+ DATE.udtFormat(trans.getIssueDate()) + "</ram:IssueDateTime>" // date
				+ buildNotes(trans)
				+ "</rsm:ExchangedDocument>"
				+ "<rsm:SupplyChainTradeTransaction>";

should be like:

		xml +=
			"<ram:GuidelineSpecifiedDocumentContextParameter>"
				+ "<ram:ID>" + getProfile().getID() + "</ram:ID>"
				+ "</ram:GuidelineSpecifiedDocumentContextParameter>"
				+ "</rsm:ExchangedDocumentContext>"
				+ "<rsm:ExchangedDocument>"
				+ "<ram:ID>" + XMLTools.encodeXML(trans.getNumber()) + "</ram:ID>";
		if (trans.getDocumentName() != null) { // !!! documentName by default is "RECHNUNG" should by null
			xml += "<ram:Name>RECHNUNG</ram:Name>";
		}
		xml +=
			"<ram:TypeCode>" + typecode + "</ram:TypeCode>"
				+ "<ram:IssueDateTime>"
				+ DATE.udtFormat(trans.getIssueDate()) + "</ram:IssueDateTime>" // date
				+ buildNotes(trans)
				+ "</rsm:ExchangedDocument>"
				+ "<rsm:SupplyChainTradeTransaction>";

best regards
franz

langfr:
Providing a document name seem to be (no longer) valid.
[...]

PR #607, opened 2024-12-08
Enable flexible PaymentReference and a DocumentName.

langfr:
The value for the PaymentRefrence was always the Invoice number. Make it settable by the user.
Business case: Payment reference should also contain the invoice date and the customer id, not only the number.

Add DocumentName for the Extended-Profile only. See #605.

Enhance a test to check the values.


jstaerk:
good suggestion thanks would like to merge maybe over the holidays I have some minutes to solve the conflict

Issue #356, opened 2024-01-12
print version of xml report


jstaerk:
Mike asked for a pdf version of the validation, simliar like the FNFE's 
[2023-17244.pdf](https://github.com/ZUGFeRD/mustangproject/files/13917909/2023-17244.pdf)

PR #612, opened 2024-12-13
#356 Added cli parameter to generate basic pdf version of xml report

Heavenfighter:
Possibillity to generate a pdf from report xml.
Needs still some refactoring since xsl has no localization and is generally simple without imports etc.


Heavenfighter:
@jstaerk  thanks for merging but someone with better xsl skills should take a look and implement localisation feature.

PR #620, opened 2024-12-19
Fix logback config

Adrian-Devries:
Fix logback config (cf. https://logback.qos.ch/codes.html#nested_if_element)


jstaerk:
what do you fix (the diff view shows all red and green), why do you fix and how do you fix it?

Adrian-Devries:
I am fixing the following: https://logback.qos.ch/codes.html#nested_if_element

Find patch attached.
[MustangProject-logback.patch](https://github.com/user-attachments/files/18209764/MustangProject-logback.patch)

PR #621, opened 2024-12-19
Fix transformation scripts

Adrian-Devries:
Fix the following warnings:

[main] WARN  c.h.x.t.LoggingTransformErrorListener - [warn] Transformation warning (net.sf.saxon.trans.XPathException: A function that computes atomic values should use xsl:sequence rather than xsl:value-of)

Warning at char 27 in xsl:value-of/@select on line 1985 column 166 
  SXWN9026: Comparison of xs:untypedAtomic? to xs:boolean will fail unless the first operand is empty

Warning at xsl:for-each on line 2083 column 78 
  SXWN9009: An empty xsl:for-each instruction has no effect

### DIFF
--- a/library/src/main/resources/stylesheets/ZUGFeRD_1p0_c1p0_s1p0.xslt
+++ b/library/src/main/resources/stylesheets/ZUGFeRD_1p0_c1p0_s1p0.xslt
@@ -1007,7 +1007,7 @@ Kosten, Verluste bzw. Schäden hätte normalerweise vorhergesehen werden können
 																											<xsl:for-each select="ram:ChargeIndicator">
 																												<xsl:for-each select="udt:Indicator">
 																													<span>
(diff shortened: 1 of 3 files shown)


jstaerk:
that should be fixed upstream and the AWV no longer maintains ZUGFeRD 1 :-(

Adrian-Devries:
Only the first commit is related to ZUGFeRD_1p0_c1p0_s1p0.xslt. The second one is related to XR_30.

Please, specify what the upstream is, so that I could address pull requests against it.

Adrian-Devries:
Pull requests have been filed against upstream projects:

- https://github.com/OpenPEPPOL/peppol-bis-invoice-3/pull/166
- https://github.com/itplr-kosit/xrechnung-schematron/pull/34

PR #624, opened 2024-12-19
Add wrapper

Adrian-Devries:
Support for multiple cycles, detecting new versions of plugins and dependencies, detecting undeclared dependencies etc.

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -1,0 +1,92 @@
+@echo off
+
+:Main
+setlocal
+setlocal enableextensions
+setlocal enabledelayedexpansion
+set "JAVA_TOOL_OPTIONS=%JAVA_TOOL_OPTIONS% -Duser.language=en-US"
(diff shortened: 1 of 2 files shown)


jstaerk:
I'm not a fan of it, it seems to be platform dependant, there seems to be some special handling of specific java versions, it's not documented nor can I see what it does. 
What do you mean by  multiple cycles and would "detecting new versions of plugins and dependencies & undeclared dependencies" not be the task of SAST tools like version eye?

Adrian-Devries:
The command script repeats compilation and test steps multiple times with different parameters (see comments inside the script), it creates log files for each single step. Find the log files attached.
[Mustang-CLI-2.15.3-SNAPSHOT-logs.zip](https://github.com/user-attachments/files/18209946/Mustang-CLI-2.15.3-SNAPSHOT-logs.zip)


In case of demand, I could provide a BASH version of it.

PR #650, opened 2025-01-07
Add net.sf.offo:fop-hyph

Adrian-Devries:
Add hyphenation patterns to avoid the following issue:

"WARN  org.apache.fop.apps.FOUserAgent - Hyphenation pattern not found."


jstaerk:
thanks

PR #668, opened 2025-01-13
Fix possible NPE when reading item charge data from xml.

### LINKED ISSUE #666: Java exception with V 2.16.0
A PDF created immediately before with 2.16.0 leads to this error during validation. Nevertheless, the validation was obviously carried out and classified as valid. The error does not occur with version 2.15.

grafik

langfr:
Should fix #666.


langfr:
Correct.

PR #674, opened 2025-01-13
Fix Java 8 issues

Adrian-Devries:
Enhanced lambda expressions etc.


jstaerk:
there seem to be unrelated changes and we also dont support java 8?!

Adrian-Devries:
> there seem to be unrelated changes and we also dont support java 8?!

The point is that the code base is still using Java 8 constructs. The pull request was meant to replace these Java 8 constructs.

PR #676, opened 2025-01-13
Fix Java 11 issues

Adrian-Devries:

PR #675, opened 2025-01-13
Fix Java 9 issues

Adrian-Devries:

PR #673, opened 2025-01-13
Fix Java 7 issues

Adrian-Devries:
Collapse catch blocks etc.

PR #672, opened 2025-01-13
Fix Java 5 issues

Adrian-Devries:
Extended for loops etc.

PR #677, opened 2025-01-13
Fix complex if constructs

Adrian-Devries:
Replace by switch


jstaerk:
fundamentally good idea but changeset obfustcated with unrelated whitespace changes etc..

Adrian-Devries:
> fundamentally good idea but changeset obfustcated with unrelated whitespace changes etc..

There are no unrelated changes. The pull request just includes previous ones (#672, #673, #674, #675, #676).

PR #701, opened 2025-01-22
Ensure Base64 decoding can handle newlines when decoding a FileAttachment

lolletsoc:


jstaerk:
never heard of that one but if I google it it seems as if you may just be right. thanks.

Issue #249, opened 2021-09-21
invalid output PDF for input with incomplete CIDsets


Roman-it20one:
Hi!

I have a valid pdf/a-1 file "pdfa1.pdf" (according to verapdf). I combine it with a valid xml "ZUGFeRD-invoice.xml" via the command line tool.

Example of command line execution:

.\jre\bin\java.exe -jar .\lib\Mustang-CLI-2.2.1.jar --action combine --source ".\temp\pdfa1.pdf" --source-xml ".\temp\ZUGFeRD-invoice.xml" --out ".\temp\combined.pdf" --format zf --version 1 --profile T

The resulting file "combined.pdf" fails pdf/a-3 validation with the following error:

"If the FontDescriptor dictionary of an embedded CID font contains a CIDSet stream, then it shall identify all CIDs which are present in the font program, regardless of whether a CID in the font is referenced or used by the PDF or not. 

2 occurrences"

Verapdf gives this additional info:

"PDCIDFont fontFile_size == 0 \|\| fontName.search(/[A-Z]{6}\+/) != 0 \|\| CIDSet_size == 0 \|\| cidSetListsAllGlyphs == true |  
root/document[0]/pages[0](9 0 obj PDPage)/contentStream[0](16 0 obj PDContentStream)/operators[51]/font[0](UCIANL+ArialMT)/DescendantFonts[0](UCIANL+ArialMT)
root/document[0]/pages[0](9 0 obj PDPage)/contentStream[0](16 0 obj PDContentStream)/operators[128]/font[0](QUBHFO+Arial-BoldMT)/DescendantFonts[0](QUBHFO+Arial-BoldMT)"

Best Regards
Roman
[files.zip](https://github.com/ZUGFeRD/mustangproject/files/7202249/files.zip)

jstaerk:
Thanks, I can reproduce it :-( I will have a look.

jstaerk:
the error 
```
Rule | Status
-- | --
Specification: ISO 19005-3:2012, Clause: 6.2.11.4, Test number: 4 |  
If the FontDescriptor dictionary of an embedded CID font contains a CIDSet stream, then it shall identify all CIDs which are present in the font program, regardless of whether a CID in the font is referenced or used by the PDF or not. | Failed
2 occurrences | Hide
PDCIDFont |  
[...]

PR #721, opened 2025-01-31
remove CIDSets from XObjects

### LINKED ISSUE #659: invalid output PDF for input with incomplete CIDsets embedded in PDXObjects
Hi,

we create PDFs in the following way: We create Invoice-PDFs (PDF/A-1b) using jasperreports. After that we combine them with an background-layer-PDF that contains the company-logo and other informations like an imprint. This 'background-PDF' is also a valid PDF/A-1b file.
Both files can be verified successfully using verapdf. We combine them using the org.apache.pdfbox.multipdf.Overlay-Class of PDFBox. The result can also be verified successfully.

When adding the XML to build a valid Zugferd-PDF using Mustang the output is not valid anymore. It has to do with the CIDSet as mentioned in issue #249 .
The problem is, that combining both PDFs using the Overlay-Class is done by wrapping the layer in an PDXObject. The patch supplied for #249 just scans for fonts in the root-dictionary.

Fonts embedded in PDXObjects are not scanned.
Replacing ZUGFeRDExporterFromA3.removeCidSet(PDDocument) with these two functions will fix the problem:

```
	private void removeCidSet(PDDocument doc) throws IOException {
		// https://github.com/ZUGFeRD/mustangproject/issues/249

		COSName cidSet = COSName.getPDFName("CIDSet");
		COSName resources = COSName.getPDFName("Resources");

		// iterate over all pdf pages
		for (Object object : doc.getPages()) {
			if (object instanceof PDPage) {

				PDPage page = (PDPage) object;
				PDResources res = page.getResources();

				// Check for fonts in PDXObjects:
				for (COSName xObjectName : res.getXObjectNames()) {
					PDXObject xObject = res.getXObject(xObjectName);
					COSDictionary d = xObject.getCOSObject().getCOSDictionary(resources);
					if (d != null) {
						PDResources xr = new PDResources(d);
						removeCIDSetFromPDResources(cidSet, xr);
					}
				}
				
				// Check for fonts in document-resources:
				removeCIDSetFromPDResources(cidSet, res);
			}
		}
	}

	private void removeCIDSetFromPDResources(COSName cidSet, PDResources res) throws IOException {
		for (COSName fontName : res.getFontNames()) {
			try {
				PDFont pdFont = res.getFont(fontName);
				if (pdFont instanceof PDType0Font) {
					PDType0Font typedFont = (PDType0Font) pdFont;

					if (typedFont.getDescendantFont() instanceof PDCIDFontType2) {
						@SuppressWarnings("unused")
						PDCIDFontType2 f = (PDCIDFontType2) typedFont.getDescendantFont();
						PDFontDescriptor fontDescriptor = pdFont.getFontDescriptor();

						fontDescriptor.getCOSObject().removeItem(cidSet);
					}
				}
			} catch (IOException e) {
				throw e;
			}
			// do stuff with the font
		}
	}
```

It would be nice, if you could integrate this patch into mustang-project.

Thanks and greetz,
Karsten

mr-mister123:
fixes #659

do you need example-files for testing?

Greetz, Karsten

### DIFF
--- a/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExporterFromA3.java
+++ b/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDExporterFromA3.java
@@ -58,6 +58,7 @@ import org.apache.pdfbox.pdmodel.font.PDCIDFontType2;
 import org.apache.pdfbox.pdmodel.font.PDFont;
 import org.apache.pdfbox.pdmodel.font.PDFontDescriptor;
 import org.apache.pdfbox.pdmodel.font.PDType0Font;
+import org.apache.pdfbox.pdmodel.graphics.PDXObject;
 import org.apache.pdfbox.pdmodel.graphics.color.PDOutputIntent;
 import org.apache.xmpbox.XMPMetadata;
 import org.apache.xmpbox.schema.AdobePDFSchema;
@@ -559,6 +560,7 @@ public class ZUGFeRDExporterFromA3 extends XRExporter implements IZUGFeRDExporte
 		// https://github.com/ZUGFeRD/mustangproject/issues/249
 
 		COSName cidSet = COSName.getPDFName("CIDSet");
+		COSName resources = COSName.getPDFName("Resources");
 
 		// iterate over all pdf pages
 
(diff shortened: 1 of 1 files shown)


jstaerk:
Closing because for me it looks like an attempt to correct something in invalid PDF-A input files, correct? As that matter is far too complicated I would not start even trying that.

mr-mister123:
> Closing because for me it looks like an attempt to correct something in invalid PDF-A input files, correct? 

No, i don't think so. According to #249 you've decided to let mustang convert pdf/a-1 to pdf/a-3 and therefore removing the cid-sets from embedded fonts with commit [#df83c30](https://github.com/ZUGFeRD/mustangproject/commit/df83c303a86d193a1b18311f045ebf63118b5aad).

> Apparently "The rules for incomplete CID sets changed between PDF/A-1a and PDF/A-3a. They are stricter in PDF/A-3a", but PDF/A-3 works completely fine without CIDsets, so the I'm now removing them.

( [see](https://github.com/ZUGFeRD/mustangproject/issues/249#issuecomment-933316935) )

The only thing i changed is, that your patch is not only applied to the fonts in the root-dictionary, but also to object that are embedded into the pdf (eg. when merging pdf-files). So for me it's just the consequent implementation of your built-in-support for converting pdf/a-1 to pdf/a-3.
Your original algorithm has not been changed, it's just consequently applied to **all** embedded fonts.

I could supply sample-files if needed.

Greetz,
Karsten

jstaerk:
I can not reproduce this issue. Please provide valid PDF/As which become invalid if the CIDs are not removed.

mr-mister123:
Here are sample-files:

[background-valid.pdf](https://github.com/user-attachments/files/19069222/background-valid.pdf)
We have an (according to vera-pdf) valid pdf used as background (valid PDF/A-1b)


[invoice-valid.pdf](https://github.com/user-attachments/files/19069223/invoice-valid.pdf)
a valid invoice-pdf (PDF/A-1b)

[invoice_with_background-valid.pdf](https://github.com/user-attachments/files/19069225/invoice_with_background-valid.pdf)
both combined to one single pdf (using pdfbox), still valid PDF/A-1b

[invoice_with_background_and_xml-invalid.pdf](https://github.com/user-attachments/files/19069224/invoice_with_background_and_xml-invalid.pdf)
and the resulting zugferd-pdf after attaching the xml using mustang-project: this results in an invalid pdf file, although the input was completely valid...

i hope this helps.

Thanks and greetz,
Karsten

PR #731, opened 2025-02-04
Got a broken translation key when visualizing XML into PDF (xr:Business_process_type)

### LINKED ISSUE #730: [cli 2.16.1] Got a broken translation key when visualizing XML into PDF (xr:Business_process_type)
Hi
I've received an XRechnung (CII) and visualized it into a PDF using Mustang CLI. The generated PDF contains a weird string that seems to be a result of broken translation.

Image

You can see the anonymized invoice here. Github doesn't allow attaching xml files, so I changed it to txt.
[xrechnung-cii.xml.txt](https://github.com/user-attachments/files/18655982/xrechnung-cii.xml.txt)

Faerballert:
PR will fix #730


jstaerk:
The PR seems to rename a key, remove it from a place where it is not used, and put a static value in another place where it *is* used but is not always correct?

Faerballert:
No, it is not just renaming. If you search for Business_process_type there are also some other places where this is in use.

There was just a conflict by using 2 different static values for same use-case

Issue #640, opened 2025-01-02
Having two items with different VATs and adding a charge/allowance will generate a invalid XML


FrankHossfeld:
It looks like that there is an issue when having an invoice with two items and different VAT. Adding a charge/allowance will generate an invalid XML.

Generate an invoice with the following two items:

* item 1: quantity: 10, price: 10, VAT 19 %
* item 2: quantity: 1, price: 10, VAT: 19 %

* invoice with two items and same VAT without a charge/allowance -> works (XML is valid)
* invoice with two items and same VAT with a charge/allowance -> works (XML is valid)
* invoice with two items (one item 19% VAT, second 7 % VAT) without a charge/allowance -> works (XML is valid)
* invoice with two items (one item 19% VAT, second 7 % VAT) with a  charge/allowance -> works (XML is **invalid** or calculation is wrong)

1. Setting up a charge with reason and percent and without VAT will result in an invalid XML
2. Setting up a charge with reason, percent and VAT will result in an valid XML but a wrong calculation

In the first case the XML for the charges look like this:
```
      <ram:SpecifiedTradeAllowanceCharge>
        <ram:ChargeIndicator>
          <udt:Indicator>true</udt:Indicator>
        </ram:ChargeIndicator>
        <ram:ActualAmount>11.00</ram:ActualAmount>
        <ram:Reason>Teuerungszuschlag</ram:Reason>
        <ram:CategoryTradeTax>
          <ram:TypeCode>VAT</ram:TypeCode>
          <ram:CategoryCode>S</ram:CategoryCode>
          <ram:RateApplicablePercent>0.00</ram:RateApplicablePercent>
        </ram:CategoryTradeTax>
      </ram:SpecifiedTradeAllowanceCharge>
```
The XML is invalid due to the tax percent are missing.

In the second case adding two charges (one for VAT 19% and a second one for 7 %) generates the following XML:
```
      <ram:SpecifiedTradeAllowanceCharge>
        <ram:ChargeIndicator>
          <udt:Indicator>true</udt:Indicator>
        </ram:ChargeIndicator>
        <ram:ActualAmount>11.00</ram:ActualAmount>
        <ram:Reason>Teuerungszuschlag</ram:Reason>
        <ram:CategoryTradeTax>
          <ram:TypeCode>VAT</ram:TypeCode>
          <ram:CategoryCode>S</ram:CategoryCode>
          <ram:RateApplicablePercent>7.00</ram:RateApplicablePercent>
[...]

PR #769, opened 2025-03-08
Draft: Fix for #640 

FrankHossfeld:
ATM this PR is nothing more than a draft. It should be used a base for a discussion how to fix the issue. Also it needs further testing!

This PR improves the TransactionCalculator-class. The calculation inside the getVATPercentAmountMap-method is improved and stored as an instance variable to avoid the calculation of the valiues several times.


FrankHossfeld:
@jstaerk no, it's not fixed in the new version. I ran the tests again and and with the new version the calculation difference is even higher.

jstaerk:
I dont yet understand why a change should be required at all, see my comments on #640

PR #805, opened 2025-04-03
Delete temp file on exit

### LINKED ISSUE #775: not deleted tmp files
Whenever I create a ZUGFeRD version of a PDF file, e.g. by combining from an existing PDF file and an XML description, the software creates tmp files in my /tmp directory, which are not deleted after the task is done. This happens with my own code and also with the CLI version.

The files look like

```
/tmp/mustang-pdf4768011548584074371.pdf
/tmp/mustang-pdf7259666362176839781.pdf
```
After digging in the source code it looks like they are created during validation of the given PDF file.

If you are a heavy user of the library, this could result in a congestion of the /tmp directory. In addition, this could be a security issue, because the files could be read by unwanted audience.

As a first solution for my use case, I delete the files with other means. But IMHO the deletion should be done by the software itself.

phax:
Make sure all temporary files are deleted at the end of the JVM session - to fix #775


jstaerk:
thanks a million, I would have thought more complicated

PR #828, opened 2025-05-05
Create a single instance of DocumentBuilderFactory / TransformerFactory.

langfr:
Create DocumentBuilderFactory / TransformerFactory with appropriate features and properties.
In PDFValidator XPathExpression had to be adjusted for attributes with namespaces.
```
<rdf:Description xmlns:fx="urn:factur-x:pdfa:CrossIndustryDocument:invoice:1p0#" rdf:about="" fx:ConformanceLevel="EXTENDED" fx:DocumentFileName="factur-x.xml" fx:DocumentType="INVOICE" fx:Version="1.0"/>
```


mr-mister123:
Be careful! Oracle states that..

>  An implementation of the DocumentBuilderFactory class is NOT guaranteed to be thread safe. It is up to the user application to make sure about the use of the DocumentBuilderFactory from more than one thread.

therefore i don't think it's a good idea to do so.
But i like the idea of centralizing the configuration of the factories. May be your Class should just create instances of BuilderFactorys that are properly configured...

greetz,
Karsten

jstaerk:
ZF2EdgeTest.testEdgeExport:329 seems to fail afterwards?

langfr:
Because of @mr-mister123's comment I wanted to change this PR.

Issue #838, opened 2025-05-20
visualizer unable to create different languages


truearken:
When using the Visualizer, the first language used to initialize the templates remains the same.

Reproduce:
```java
var visualizer = new ZUGFeRDVisualizer();
visualizer.visualize(content, Language.DE) // --> results is in german
visualizer.visualize(content, Language.EN) // --> result is still in german
```

This occurs because of the null checks in the initTemplates() Method.

langfr:
Created PR #843.

PR #843, opened 2025-05-23
Re-Initialize the HTML-template on language change.

langfr:
See #838.


truearken:
while this works it would probably be bad for the user since a language change will always result in a new compilation. i think it'd be better to cache every compiled sheet so a language change won't take too much time.

langfr:
Changed fix to use a language-based Map of templates.
There are only 3 different languages, so the map will contain at max 3 templates.
Constantly changing between one of these languages will be more lightweight.

truearken:
looks really nice!

truearken:
@jstaerk This is merged already, but I noticed that this still wouldn't work.
The null check for the HashMap in the initTemplates() method denies the code to ever run into the case where a language isn't initialized yet. Was this even tested? The tests seem to always generate a new object.

The code should look like this:
```java
if (mXsltHTMLTemplates == null) {
	mXsltHTMLTemplates = new HashMap<>();
}
if (mXsltHTMLTemplates.get(lang) == null) {
        Templates mXsltHTMLTemplate = mFactory.newTemplates(new StreamSource(CLASS_LOADER.getResourceAsStream(RESOURCE_PATH + "stylesheets/xrechnung-html." + lang.name().toLowerCase() + ".xsl")));
        mXsltHTMLTemplates.put(lang, mXsltHTMLTemplate);
}
```

PR #898, opened 2025-07-23
fix: fix MethodNotFoundException when using jackson 2.19.2

mosesonline:
Projects with jackson 2.19.2 XMLMapper throw exception MethodNotFoundException ParserMinimalBase.<init>


jstaerk:
You say that this exception is thrown in 2.19.2? Then why should we migrate to this version?

langfr:
We're using version 2.22.1 currently.

PR #899, opened 2025-07-24
fix encoding error in ZUGFeRDInvoiceImporter.java

asdf00000:
fix encoding error when JVM has a standard encoding other than UTF-8


jstaerk:
closes #900

Issue #950, opened 2025-10-18
2.19.1: ClassNotFoudException while running the PDFValidator


FrankHossfeld:
Updating to the latest release (2.19.1) I ran into some trouble. 

Running the `PDFValidator` I get a CNFE saying:

```
java.lang.NoClassDefFoundError: Could not initialize class org.mustangproject.validator.PDFValidator
	at org.mustangproject.validator.ZUGFeRDValidator.internalValidate(ZUGFeRDValidator.java:100)
	at org.mustangproject.validator.ZUGFeRDValidator.validate(ZUGFeRDValidator.java:276)
...
Caused by: java.lang.ExceptionInInitializerError: Exception java.lang.NoClassDefFoundError: org/verapdf/pdfa/flavours/PDFAFlavour [in thread "main"]
	at org.mustangproject.validator.PDFValidator.<clinit>(PDFValidator.java:57)
	at org.mustangproject.validator.ZUGFeRDValidator.internalValidate(ZUGFeRDValidator.java:100)
	at org.mustangproject.validator.ZUGFeRDValidator.validate(ZUGFeRDValidator.java:276)
```

Might be related to the fact that the verapdf-classes are not included in the shaded artifact.

FrankHossfeld:
yep, you are right, I missed to add the classifier for the validator dependency. Adding it, solve the issue. Maybe, the release notes should point that (or didi I missed it?)

Please, can yo explain, why the mustangproject starts using the shade-plugin with 2.19.1?

jstaerk:
@FrankHossfeld you are right, thank you, switched to <shadedArtifactAttached>false</shadedArtifactAttached>

PR #957, opened 2025-10-22
fix for #950

FrankHossfeld:
This PR fixes the issue #950 & #227.

Changes:

* add dependency management (parent pom)
* add plugin management (parent pom)
* remove all version-tags from dependencies (now managed inside the dependencyManagement and pluginManagement) in the child pom (library & validator & cli)
* remove the maven-shade-plugin (library-module-pom)
* remove the maven-shade-plugin (validator-module-pom)
* add main class in the maven-shade-plugin of the cli pom
* add two dependencies to avoid the CNFE while running the cli (https://github.com/ZUGFeRD/mustangproject/issues/227)

After merging, this shaded library is no longer needed in core & validator


FrankHossfeld:
@jstaerk Sure, it is. I'll need to update the PR first. Then, in case you are interested, we can try to merge it.

_"so far the maven central release process was not very stable towards"_ ... that sounds like the repo needs a better CI/CD. Looking at the things GitHub offers, there are a multitude of possibilities to improve CI/CD. And it does not take much. So, why not improve the branch concept, use some actions and create an automated build and deploy for SNAPSHOT? What do you think? (Of course, this will be a new issue or two ...)

In my opinion, the shaded plugin creates a lot of problems. Maybe that's the reason why you are running into.

FrankHossfeld:
@jstaerk Due to changes related the use of Maven Central (https://central.sonatype.org/publish/maven-central-publishing-limits/) it might also be a good idea dropping shaded jars.

jstaerk:
can you send me an email and we arrange a short meeting to talk? thanks

FrankHossfeld:
Done ... [linkedin]

FrankHossfeld:
@jstaerk Do you have an Discord account?

jstaerk:
> @jstaerk Do you have an Discord account?

no but i sent you a mail a while ago

FrankHossfeld:
yeah, just found it ...  landed among the spam ...

answered your email ...

PR #964, opened 2025-11-06
Use IZUGFeRDExportableItem.getCalculation instead of LineCalculator constructor

WorkProg:


WorkProg:
@jstaerk, sorry to bother you again. Do you know if this can be merged soon? We are currently blocked by this in our E-Invoicing integration.

WorkProg:
@jstaerk, this requires us to currently copy the whole of the xml building into our code base just because we cannot control how calculation works. It would be good to know if we can expect this to be merged at some point.

Kind regards Roland.

timo-rohwedder:
@jstaerk +1

jstaerk:
There should be no need to control how calculation works because the norm specifies exactly one way to calculate?

WorkProg:
The Problem is, that we want to use the same sums everywhere to avoid there being a problem where we display one value to a user in the GUI and there is a difference in rounding or something like, and we end up with a different value in the XRechnung, especially if the PDF still shows the value displayed in the GUI making this difficult to spot.

We would rather if there is an error in aggregation that it is consistent in all different representations of an invoice rather than the XRechnung being correct for some reason, but different to everything else.

I would argue that if this is not merged you should instead remove the function `getCalculation` as currently if you do override it you get an invalid document were row sums would not match the total sum.

IIRC the cases that ran into problems for us are those where the unit price is calculated from the row total i.e. you have dynamic pricing and the unit price is the average price. Due to a maximum precision for the unit price in our system this results in rounding discrepancies when Mustang does the summation.

Issue #993, opened 2025-12-17
Cash discount not parsed


exactt:
Hi,
thanks again for this great library!

I have an issue with cash discounts not being parsed when they are provided in a certain format:

`<ram:SpecifiedTradePaymentTerms>`
`        <ram:Description>Cash Discount</ram:Description>`
`        <ram:DueDateDateTime>`
`	         <udt:DateTimeString format="102">20251226</udt:DateTimeString>`
`	 </ram:DueDateDateTime>`
`	 <ram:ApplicableTradePaymentDiscountTerms>`
`	           <ram:BasisAmount>135.64</ram:BasisAmount>`
`		   <ram:CalculationPercent>2</ram:CalculationPercent>`
`		   <ram:ActualDiscountAmount>2.71</ram:ActualDiscountAmount>`
`	</ram:ApplicableTradePaymentDiscountTerms>`
` </ram:SpecifiedTradePaymentTerms>`

I placed this snippet inside a test file which was provided in #692 : [response_1737188825551_ext_skonto.txt](https://github.com/user-attachments/files/24219153/response_1737188825551_ext_skonto.txt)
Here only the description of the cash discount is parsed. Not the other fields.

Is this expected behaviour?

exactt:
The part I commented out is from the original file. You can ignore this.

ThomasChr:
Hi @exactt!

First of all: I'm a little bit puzzled by this invoice. The Total Sum is 529.87 € but the 4% Skonto are applied for 8721.32 € which doesn't seem to be correct at all...
How has this invoice been generated?

Also I was not able to find any tool to parse this correctly and display the discount amount. What are you using?

tldr: A little bit more info would be great :)

Thomas

ThomasChr:
Thanks!
Do you any tool at hand which can show the discount for me to cross-check?

exactt:
Quba Viewer seems to get it:

<img width="542" height="306" alt="Image" src="https://github.com/user-attachments/assets/a9f39548-47d4-4099-a848-74c0c452e70c" />

ThomasChr:
Yeah. It does display it, but there is no further calculation with it.
The text 'Bis zum 15.02.2026 erhalten Sie 4,000  % Skonto, 21.19 EUR' comes straight from the XML.

I will dig into MustangLibrary and see what I can find.

exactt:
I tested your code and it worked for both of my two real-world invoices (from different suppliers). Cash discounts were properly parsed.

[...]

PR #1015, opened 2026-01-21
fix for #993

ThomasChr:


ThomasChr:
Since this fix seems to have no side effects, looks good to me, and is used in production - I think we can merge it!
Reference: https://github.com/ZUGFeRD/mustangproject/issues/993#issuecomment-3889398966

PR #1037, opened 2026-02-11
Compress attachments

pgjacobs:
Currently, PDFBox is used to attach the XML file and other attachments without filters.

In particular, the size of the sometimes very large files factur-x.xml/xrechnung.xml would be drastically reduced if they were compressed when attached, e.g., using FLATE_DECODE. The resulting PDF would still comply with the PDF/A-3 standard, in which this is generally permitted, and should therefore satisfy the ZUGFeRD/X-Rechnung guidelines.

Since I don't see any disadvantage in using it, I suggest always attaching all attachments in compressed form.


erras-it:
Would it make sense to make compression configurable? In particular, to be able to maintain the existing behaviour.

pgjacobs:
Hi @erras-it,
I added the configuration option `org.mustangproject.ZUGFeRD.IZUGFeRDExporter#setEnablePDFAttachmentCompression` defaulting to false, so the existing behaviour is retained by default.

jstaerk:
thanks! the disadvantage is that sometimes people tend to expect certain values in the binary file, i.e. in an uncompressed part (fnfe validator I'm looking in your direction, cough) but if you make it configurable like you did there should indeed be no issue

PR #1087, opened 2026-04-08
Add missing quotation marks and revert corruption of pdfs.

langfr:


langfr:
The pdf files itself which were part of the previous commit are no longer readable by Acrobat Reader, MS Edge or PDF24.
Acrobat reader says
image
others are silent and show only an empty page.
The prior versions where readable.
This leads to test errors then because of the invalid pdf, not the (partly) expected  errors in the included invoice xml.

Example: MustangBeispiel20221026.pdf

langfr:
Obsolete now.

langfr:
.

Issue #956, opened 2025-10-22
Reading an invoice with multiple SpecifiedTradePaymentTerms only returns the last description


PhilippGoetje:
I have an invoice with the following content. It validates and to my knowledge is in accordance with the ZF extended standard.

`
<ram:SpecifiedTradePaymentTerms>
<ram:Description>#SKONTO#TAGE=10#PROZENT=3.0#BASISBETRAG=2141.56# </ram:Description>
<ram:DueDateDateTime>
<udt:DateTimeString format="102">20251024</udt:DateTimeString>
</ram:DueDateDateTime>
<ram:ApplicableTradePaymentDiscountTerms>
<ram:BasisAmount>2141.56</ram:BasisAmount>
<ram:CalculationPercent>3.0</ram:CalculationPercent>
</ram:ApplicableTradePaymentDiscountTerms>
</ram:SpecifiedTradePaymentTerms>
<ram:SpecifiedTradePaymentTerms>
<ram:Description>Bis zum 13.11.2025 ohne Abzug</ram:Description>
<ram:DueDateDateTime>
<udt:DateTimeString format="102">20251113</udt:DateTimeString>
</ram:DueDateDateTime>
</ram:SpecifiedTradePaymentTerms>
`

Mustang only returns the description of the second SpecifiedTradePaymentTerms item. 
`
"currency": "EUR",
"paymentTermDescription": "Bis zum  13.11.2025 ohne Abzug",
"issueDate": "2025-10-14T00:00:00.000+00:00",
`

This makes parsing of the Skonto impossible.

PR #1094, opened 2026-04-14
Fix: import all SpecifiedTradePaymentTerms blocks, not just the last …

PhilippGoetje:
# Fix reading invoices with multiple SpecifiedTradePaymentTerms (#956)

**Target branch:** `ZUGFeRD/mustangproject:master`  
**Branch:** `PhilippGoetje/mustangproject:fix/multiple-payment-terms-import`  
**Closes:** #956

---

## What and Why

When a CII invoice contains more than one `<ram:SpecifiedTradePaymentTerms>` block
(permitted in the ZUGFeRD Extended / Factur-X Extended profile with cardinality 0..n),
`ZUGFeRDInvoiceImporter` silently discarded all but the last description. It also
failed to parse `#SKONTO#` cash-discount text from any block after the first.

**Root causes — two independent bugs:**

### Bug 1 — DOM loop overwrites Description on every block

```java
// Before (lines 852-853 in extractInto): called once per block, last wins
zpp.setPaymentTermDescription(paymentTermChilds.item(i).getTextContent());
```

Each `SpecifiedTradePaymentTerms` node was visited in document order, but
`setPaymentTermDescription` overwrote the previous value, so only the last
`Description` survived.

### Bug 2 — `potentialCashDiscountTerms` only receives the first Description

```java
// Before (line 666): XPath string() returns first node in document order only
potentialCashDiscountTerms = extractString(
  "//*[local-name()=\"SpecifiedTradePaymentTerms\"]/*[local-name()=\"Description\"]");
```

When the first block contained a plain-text due date and the second block carried
the `#SKONTO#` string, no cash discount was ever parsed.

---

### New file: `library/src/main/java/org/mustangproject/PaymentTerms.java`

A public, mutable implementation of `IZUGFeRDPaymentTerms` (description, due date,
discount terms). Previously the only implementation was a private inner class inside
a test. The importer uses this class to materialise each payment-terms block;
callers building invoices programmatically can use it too.

### `ZUGFeRDInvoiceImporter.java`

**DOM loop (`extractInto`, settlement section):** For each `SpecifiedTradePaymentTerms`
node, a fresh `PaymentTerms` object is created, populated with the block's
`Description`, `DueDateDateTime`, and `DirectDebitMandateID`, then added via
`Invoice.addPaymentTerms()`. The per-block due date is also propagated to the
top-level `dueDate` variable (last value wins — existing behaviour preserved).

**`potentialCashDiscountTerms` (CII branch):** Changed from a single
`extractString(...)` call to a nodeset XPath that collects **all** `Description`
texts and joins them with `\n`. This feeds the existing `#SKONTO#` regex parser
which already iterates lines — no changes to that parser needed.

**Backward compatibility:** `getPaymentTermDescription()` still returns the first
description string (set via the unchanged XPath call at line 823). Code that only
reads this single string continues to work.

### `Invoice.java`

Added `@JsonIgnore` to the `paymentTerms` ArrayList field and the
`getPaymentTerms()` / `getExtendedPaymentTerms()` methods. This keeps the
JSON serialisation schema identical to the pre-fix schema — the structured list
is a runtime API, not a JSON serialisation feature; the `paymentTermDescription`
string already covers that.

### `CashDiscountFacturXTest.java` — two new test methods

| Test | What it covers |
|---|---|
| `testMultiplePaymentTermsExtendedList` | 3-block invoice: asserts `getExtendedPaymentTerms()` returns all 3 terms with correct descriptions and due dates; asserts backward-compat `getPaymentTermDescription()` returns the first |
| `testMultipleSkontoCashDiscountsFromDescriptions` | 2-block invoice matching the screenshot in #956: both blocks use `#SKONTO#` text convention; asserts 2 structured terms, first description returned by `getPaymentTermDescription()`, and **2 CashDiscount objects** parsed (3% / 8 days and 0% / 10 days) |

---

## Specification Alignment

Multiple `<ram:SpecifiedTradePaymentTerms>` blocks are explicitly allowed at
cardinality 0..n in the UN/CEFACT SCRDM CII D16B schema underlying both ZUGFeRD
Extended and Factur-X Extended. The pattern — one skonto block plus one net-due
block — is the standard German practice for encoding early-payment discounts
alongside a fallback net payment term.

---

## Test Results

Full library test suite: **156 tests, 0 failures, 0 errors** (run on Java 22 /
mustangproject 2.22.1-SNAPSHOT).

The two pre-existing failures mentioned in other PRs (`LibraryTest.testLibraryPushAllowances`,
`PDFValidatorTest.testPDFValidation`) are unrelated to payment terms and appear
identically before and after this change.

---

## Changed Files

- `library/src/main/java/org/mustangproject/PaymentTerms.java` *(new)*
- `library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDInvoiceImporter.java`
- `library/src/main/java/org/mustangproject/Invoice.java`
- `library/src/test/java/org/mustangproject/ZUGFeRD/CashDiscountFacturXTest.java`


ichbimtul:
Hi,

I originally encountered this bug and forwarded it to Philipp, who then created this PR.

This issue has been affecting our workflow quite significantly for about 6 months now. In our case, roughly 50% of incoming invoices are currently parsed without discount (Skonto) information, even though it is present. This leads to incorrect data processing and requires manual correction.

It would be greatly appreciated if someone could take a look at this PR when time permits. I’m happy to provide additional details or examples if that helps.

PR #468, opened 2024-09-09
Fix validator dependencies

J-N-K:
Without these dependencies the `validator` project cannot be build. The build still fails due to a validation error in a test, though.

```
[INFO] --- compiler:2.3.2:compile (default-compile) @ validator ---
[INFO] Compiling 10 source files to /Users/jan/git/mustangproject/validator/target/classes
[INFO] -------------------------------------------------------------
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[23,28] error: package org.apache.commons.io does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[24,16] error: package org.dom4j does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[25,16] error: package org.dom4j does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[26,19] error: package org.dom4j.io does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[27,19] error: package org.dom4j.io does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[112,14] error: cannot find symbol
[ERROR]  class ZUGFeRDValidator
/Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[285,2] error: cannot find symbol
[ERROR]  class ZUGFeRDValidator
/Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[285,24] error: cannot find symbol
[ERROR]  class ZUGFeRDValidator
/Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[287,11] error: package org.dom4j does not exist
[ERROR] /Users/jan/git/mustangproject/validator/src/main/java/org/mustangproject/validator/ZUGFeRDValidator.java:[289,14] error: cannot find symbol
[ERROR]  class ZUGFeRDValidator
[...]

PR #1105, opened 2026-04-27
[OPTIONAL] Add defensive warning for line-level allowance/charge arithmetic (BT-137/BT-142)

PhilippGoetje:
# PR-2 [OPTIONAL]: Add defensive warning for line-level allowance/charge arithmetic (BT-137/BT-142)

**Branch:** `fix/validator-line-allowance-charge-arithmetic`
**Target:** `master`

> **This PR is explicitly marked optional.** It can be merged, dropped, or deferred
> independently of any other changes. See `## Why this PR is marked optional` below.

---

## Summary

- **Optional, defensive validator warning.**
- For every line-level `<ram:SpecifiedTradeAllowanceCharge>` that carries a `<ram:CalculationPercent>` element, the validator emits a **warning** if `|ActualAmount − BasisAmount × CalculationPercent / 100| > 0.02`.
- Severity is **warning** only. The validation summary status (`valid`/`invalid`) is **not affected**.

## Why this PR is marked optional

- The check is **not** in EN 16931-1:2017+A1:2019 and **not** in CEN schematron v1.3.15.
- EN 16931-1:2026 (published 2026-03-31) confirms **BR-67** (line-net formula, [issue #445](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/445)) but does **not** directly mandate `ActualAmount = BasisAmount × CalculationPercent / 100` for each individual allowance/charge entry.
- TC 434 chair Paul Simons committed a **BT-131 calculation rule** for EN 16931-1:2026 ([issue #350](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/350)), but explicitly stated this is for the line total, not for individual BG-27/BG-28 arithmetic identity.
- The April 2026 CEN/TC 434 schematron release ([issue #468](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/468)) **may** add such a rule. If it does, this internal warning can be removed in the same PR that bumps the schematron version.
- The warning is anchored on the **semantic definitions** of BT-137 / BT-142 in EN 16931-1:2017+A1:2019 (normative today): *"the base amount used in conjunction with the percentage to calculate the allowance/charge amount."* A document that provides a `BasisAmount` and `CalculationPercent` but whose `ActualAmount` deviates materially violates this semantic intent.

## Scope of change

| File | Change |
|------|--------|
| `validator/src/main/java/org/mustangproject/validator/XMLValidator.java` | New private `checkLineAllowanceChargeArithmetic()` method; called at the end of `checkArithmetics()`. Uses the existing `DocumentBuilder`/`XPath` pattern already present in the class. |
| `validator/src/test/java/.../XMLValidatorAllowanceChargeArithmeticTest.java` | Two tests: (1) identity holds within 0.02 → no BT-137/BT-142 warning; (2) deviation = 3.565 → warning fires, message contains `BT-137`/`BT-142`. |
| `validator/src/test/resources/validLineLevelAllowanceArithmetic.xml` | Minimal Extended-profile CII fixture with correct arithmetic (BasisAmount=64.25, Pct=10, ActualAmount=6.42). |
| `validator/src/test/resources/invalidLineLevelAllowanceArithmetic.xml` | Same fixture with intentionally wrong ActualAmount=9.99 (deviation 3.565). |

## Validation / test plan

- [x] `mvn -pl validator -am test` green.
- [x] `testNoWarningWhenArithmeticIdentityHolds` — valid fixture produces no BT-137/BT-142 warning.
- [x] `testWarningFiredWhenArithmeticIdentityViolated` — invalid fixture produces exactly one warning with message containing `BT-137`/`BT-142`.
- [x] All existing validator tests continue to pass without regressions.

## Issues addressed

- Related to **[ConnectingEurope/eInvoicing-EN16931 #350](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/350)** — TC 434 chair commits a BT-131 calculation rule for EN 16931-1:2026; a per-allowance/charge identity is **not** part of that commitment.
- Related to **[ConnectingEurope/eInvoicing-EN16931 #445](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/445)** — BR-67 line-net formula confirmed for EN 16931-1:2026.
- Related to **[ConnectingEurope/eInvoicing-EN16931 #468](https://github.com/ConnectingEurope/eInvoicing-EN16931/issues/468)** — next CEN/TC 434 schematron release planned April 2026.
- Related to **#942** — same `checkArithmetics` family of warnings; jstaerk's closing comment there explicitly references CEN issue #350.

## Notes

- **Independent** from the line-level `BasisAmount` export fix (`fix/allowance-charge-percent`); separate branch off master; no shared commits.
- **Easy to revert** as a single commit if the upcoming CEN/TC 434 schematron adds an equivalent rule. Trigger conditions are documented in `.project/CEN-EN16931-tracking.md`.
- The 0.02 slack is consistent with the existing arithmetic checks in `checkArithmetics()` and accounts for legitimate rounding at 2 decimal places.

### DIFF
--- a/validator/src/main/java/org/mustangproject/validator/XMLValidator.java
+++ b/validator/src/main/java/org/mustangproject/validator/XMLValidator.java
@@ -464,6 +464,94 @@ public class XMLValidator extends Validator {
 			LOGGER.error(e.getMessage(), e);
 		}
 
+		checkLineAllowanceChargeArithmetic(context);
+	}
+
+	/**
+	 * Checks that for every line-level SpecifiedTradeAllowanceCharge with a CalculationPercent,
+	 * the identity ActualAmount = BasisAmount * CalculationPercent / 100 holds within 0.02.
+	 *
+	 * This is an informational warning anchored on the semantic definitions of BT-137 / BT-142
+	 * in EN 16931-1:2017+A1:2019: "the base amount used in conjunction with the percentage to
+	 * calculate the allowance/charge amount."
+	 *
+	 * No normative BR-* enforces this identity today (CEN explicitly left it out of scope in
+	 * ConnectingEurope/eInvoicing-EN16931 issue #350). This warning will become redundant if a
+	 * future CEN/TC 434 schematron release adds such a rule (tracked in .project/CEN-EN16931-tracking.md).
+	 */
+	private void checkLineAllowanceChargeArithmetic(ValidationContext context) {
+		try {
+			final DocumentBuilder db = XMLTools.getDocumentBuilder(true);
+			final Document doc = db.parse(new InputSource(new StringReader(zfXML)));
+			final XPath xpath = XPathFactory.newInstance().newXPath();
+
+			final NodeList lineItems = (NodeList) xpath.compile(
+				"//*[local-name()='IncludedSupplyChainTradeLineItem']"
+			).evaluate(doc, XPathConstants.NODESET);
+
+			for (int i = 0; i < lineItems.getLength(); i++) {
+				final Node lineItem = lineItems.item(i);
+
+				final NodeList allowanceCharges = (NodeList) xpath.compile(
+					"*[local-name()='SpecifiedLineTradeSettlement']" +
+					"/*[local-name()='SpecifiedTradeAllowanceCharge']"
+				).evaluate(lineItem, XPathConstants.NODESET);
+
+				for (int j = 0; j < allowanceCharges.getLength(); j++) {
+					final Node ac = allowanceCharges.item(j);
+
+					final String pctStr = (String) xpath.compile(
+						"*[local-name()='CalculationPercent']"
+					).evaluate(ac, XPathConstants.STRING);
+					if (pctStr == null || pctStr.isBlank()) {
+						continue;
+					}
+
+					final String basisStr = (String) xpath.compile(
+						"*[local-name()='BasisAmount']"
+					).evaluate(ac, XPathConstants.STRING);
+					final String actualStr = (String) xpath.compile(
+						"*[local-name()='ActualAmount']"
+					).evaluate(ac, XPathConstants.STRING);
+					final String chargeIndicatorStr = (String) xpath.compile(
+						"*[local-name()='ChargeIndicator']/*[local-name()='Indicator']"
+					).evaluate(ac, XPathConstants.STRING);
+
+					if (basisStr == null || basisStr.isBlank()
+							|| actualStr == null || actualStr.isBlank()) {
+						continue;
+					}
+
+					final BigDecimal basis = new BigDecimal(basisStr.trim());
+					final BigDecimal pct = new BigDecimal(pctStr.trim());
+					final BigDecimal actual = new BigDecimal(actualStr.trim());
+					final BigDecimal expected = basis.multiply(pct)
+						.divide(BigDecimal.valueOf(100), 18, java.math.RoundingMode.HALF_UP);
+					final BigDecimal deviation = actual.subtract(expected).abs();
+					final boolean isCharge = "true".equalsIgnoreCase(
+						chargeIndicatorStr != null ? chargeIndicatorStr.trim() : "");
+
+					if (deviation.compareTo(new BigDecimal("0.02")) > 0) {
+						context.addResultItem(new ValidationResultItem(ESeverity.warning,
+							"Line " + (i + 1) + " " + (isCharge ? "charge" : "allowance")
+							+ ": ActualAmount " + actual
+							+ " differs from BasisAmount * CalculationPercent / 100"
+							+ " (= " + expected.setScale(2, java.math.RoundingMode.HALF_UP) + ")"
+							+ " by more than 0.02."
+							+ " EN 16931-1:2017+A1:2019 BT-137/BT-142 define BasisAmount as"
+							+ " the base amount used in conjunction with the percentage to"
+							+ " calculate the allowance/charge amount."
+						).setSection(10));
+					}
+				}
+			}
+		} catch (IrrecoverableValidationError e) {
+			LOGGER.error(e.getMessage(), e);
+		} catch (XPathExpressionException e) {
+			LOGGER.error(e.getMessage(), e);
+		} catch (Exception e) {
+			LOGGER.error("checkLineAllowanceChargeArithmetic failed: " + e.getMessage(), e);
+		}
 	}
 
 	/***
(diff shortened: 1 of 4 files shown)


Paula-Agent:
Reviewed this because it has been open since April with no feedback. I could not build or run anything — no Java runtime is available to me — so everything below is read from source, from the released CEN artefacts and from the schematron this repository bundles. Please treat the statements about *behaviour* as things to check, not as test results.

**Short version:** the premise holds, and it is on firmer ground now than when you wrote it. But the tolerance is justified by a sentence that does not match the code, and three details would misreport on documents that are valid.

## 1. Your own trigger condition has fired, and the answer is "no"

You wrote that the April 2026 CEN/TC 434 release *may* add an equivalent rule, and that the warning could then be removed. That release shipped: `validation-1.3.16`, 2026-04-13 — two weeks before this PR was opened. I diffed the rule identifiers against `validation-1.3.15`:

* `cii/schematron/CII/EN16931-CII-model.sch` — nothing added; `BR-CO-25` and `BR-CO-27` removed (UBL model: `BR-CO-25` removed).
* `cii/schematron/CII/EN16931-CII-syntax.sch` — 28 rules added, `CII-SR-467` through `CII-SR-494`. Every one of them is a cardinality or consistency test (`count(…) <= 1` and the like). None is an arithmetic identity.

So there is still no rule of the shape `ActualAmount = BasisAmount × CalculationPercent / 100` in the current artefacts. Your "optional, may become redundant" framing is now a fact rather than a forecast, which removes the main reason to defer the PR. Worth saying so in the description.

## 2. The 0.02 is justified by something that isn't there

> The 0.02 slack is consistent with the existing arithmetic checks in `checkArithmetics()`.

There is no 0.02 in `checkArithmetics()`. The only numeric comparison it reaches is in `checkSubInvoiceLineHierarchy`, and it is exact — `childSum.compareTo(groupTotal) != 0`. The rest of the method only catches `ArithmeticException` coming out of the importer.

The two anchors that do exist point at a larger number:

* The XRechnung schematron bundled here (`validator/src/main/resources/xslt/XR_30/XRechnung-CII-validation.xslt`) defines the slack as `if($documentCurrencyCode = 'HUF') then 0.5 else 0.02`. 0.02 is right for most currencies, but it is deliberately currency-dependent — a hard-coded literal will warn on HUF invoices where the surrounding schematron deliberately does not.
* More to the point: the one normative rule of exactly this shape, an amount derived from a base times a percentage, is **BR-CO-17** (VAT category tax amount from `BasisAmount × RateApplicablePercent / 100`). It allows **± 1**, not ± 0.02:

```
abs(CalculatedAmount) - 1 <= round(abs(BasisAmount) * (Rate div 100) * 10 * 10) div 100
  and abs(CalculatedAmount) + 1 >= round(abs(BasisAmount) * (Rate div 100) * 10 * 10) div 100
```

Line-level allowances are typically smaller amounts than VAT category totals, so the relative rounding room there is larger, not smaller. And in CEN issue #350, which you cite, @csautereau states the reason the line-level rule was dropped in the first place: *"It has been withdrawn because of rounding issues for too much companies."* A warning fifty times tighter than the norm's own tolerance for the same kind of computation looks likely to reproduce exactly that complaint. Reusing the `slackValue` expression rather than a literal would be both defensible and cheap.

## 3. Multiple `ActualAmount` is not hypothetical, and the code silently takes the first

`XPathConstants.STRING` on a node-set yields the string value of the first node in document order. The CII D16B XSD permits `ram:ActualAmount` — and likewise `ram:BasisAmount` and `ram:CalculationPercent` — 0..n inside `ram:SpecifiedTradeAllowanceCharge`. A document carrying two amounts is therefore compared against one of them with nothing to indicate it.

This case is real: CEN issue #490 (2026-04-07) was filed because the CII XSLT *failed* on such a document, and 1.3.16 answered it with `CII-SR-473` (`count(ram:ActualAmount) <= 1`, fatal), whose context is `//ram:SpecifiedTradeAllowanceCharge` — line level included. But this repository bundles the 1.3.14.2 artefacts (`xslt/en16931schematron/EN16931-CII-validation.xslt` carries `version 1.3.14.2 - Last update: 2025-06-15`), which contain neither that rule nor the fix. Until the bundle is bumped, such a document passes the schematron here and lands in this Java code.

Cheapest remedy: skip the entry when `count(…) > 1`, or make the index explicit — CEN's own BR-CO-11 and BR-CO-12 write `ram:ActualAmount[1]` for precisely this reason.

## 4. One unparseable number silences the rest of the invoice

The `try` wraps both loops, so a `NumberFormatException` out of `new BigDecimal(basisStr.trim())` on the first line abandons the check for every later line. It is logged, not reported, so the outcome is indistinguishable from "everything fine". Moving the three conversions into a per-entry `try` keeps the remainder of the document covered. This is the same family of problem being argued about in #1159 and #1160.

## 5. Smaller points

* The message says `"Line " + (i + 1)`, which is the position in the document. What the recipient sees on the invoice is BT-126, `ram:AssociatedDocumentLineDocument/ram:LineID`. That is the number someone will search for.
* UBL is silently out of scope. As far as I can read it, the `checkArithmetics` call in `validate()` sits outside the `"CII".equals(context.getFormat())` block and therefore runs for UBL as well, where the XPath matches nothing at all. The same semantics live there as `cac:InvoiceLine/cac:AllowanceCharge` with `cbc:BaseAmount`, `cbc:MultiplierFactorNumeric` and `cbc:Amount`. A Javadoc line saying "CII only" would save the next reader the search.
* `xpath.compile(…)` sits inside the inner loop, so the same four expressions are recompiled once per allowance/charge, on top of a second full DOM parse of the document. Hoisting them out costs nothing to write.
* The Javadoc says CEN "explicitly left it out of scope in ... issue #350". #350 was not closed as out of scope — @midran closed it on 2025-02-24 by forwarding it to CEN/TC 434, and @SimonsPaul wrote only *"A validation rule for BT-131 will be added in the Revised EN16931-1"*, without saying anything either way about BG-27/BG-28. Since a code comment outlives a PR description, the two quotations above would serve it better.

## 6. It no longer applies

`mergeable_state` is `dirty`. The hunk anchors on the tail of `checkArithmetics`, which master has since reformatted: `catch (XPathExpressionException e)` and `catch (ParseException e)` were merged into `catch (ParseException | XPathExpressionException e)`, and the blank line before the closing brace is gone. The new method itself still applies cleanly; only the single call line has to be placed again.

One thing aged in your favour. `disableArithmeticCheck()` landed after your base (#1170). Because your call sits inside `checkArithmetics`, the new check inherits that opt-out for free — worth a line in the description, since "there is no way to switch it off" would otherwise be the obvious objection to a heuristic warning.

---

Provenance, so it can be weighed: rule identifiers were diffed between the `validation-1.3.15` and `validation-1.3.16` tags of the CEN artefacts repository; BR-CO-17 and `slackValue` were read out of the schematron files themselves; nothing was compiled or executed. The reasoning about false positives at line level comes from having implemented the document-level BR-CO-11/12/13 checks against the KoSIT test corpus, not from running yours.

Issue #1001, opened 2025-12-26
CTC-FR (Flux2 CII v1.2.0): cannot generate mandatory ExchangedDocument/IncludedNote SubjectCode PMT/PMD/AAB


meparis:
Problem

Mustang can generate ram:IncludedNote, but the subjectCode field is typed as SubjectCode (an enum). The enum currently does not expose all required UNTDID 4451 values.

As a result, it is not possible to generate the three subject codes required by the French CTC-FR validation rules:

PMT (exists in the enum)

PMD (missing)

AAB (missing)

This prevents producing Factur-X invoices that pass CTC-FR validation, even though these codes are valid UNTDID 4451 values and allowed by Factur-X.

Expected behavior

Mustang should allow setting ram:IncludedNote/ram:SubjectCode to at least:

PMT

PMD

AAB

so that a caller can generate the mandatory French legal notes.

Actual behavior

SubjectCode enum is missing PMD and AAB, so the API cannot produce these values (unless users fork/patch or bypass the enum).

References

Factur-X 1.08: BT-21 is based on UNTDID 4451 subject code list (CII ram:IncludedNote/ram:SubjectCode).

Factur-X 1.08 codedb includes AAB, PMD, PMT as allowed enumerations.

French CTC-FR Schematron (Flux 2, CII) requires presence/uniqueness of notes with SubjectCode = PMT / PMD / AAB (BR-FR-05 / BR-FR-06). (local file: 20251114_BR-FR-Flux2-Schematron-CII_V1.2.0.sch)

Proposed fix

Option A (minimal, backward compatible):

Add AAB and PMD constants to org.mustangproject.SubjectCode.

Option B (more future-proof):

Allow a free-text / custom subject code (String) in IncludedNote, keeping the enum as a convenience wrapper.

Acceptance criteria

A user can create 3 IncludedNote entries using AAB, PMD, PMT.

Generated XML contains:

<ram:IncludedNote><ram:SubjectCode>AAB</ram:SubjectCode>…</ram:IncludedNote>

<ram:IncludedNote><ram:SubjectCode>PMD</ram:SubjectCode>…</ram:IncludedNote>

<ram:IncludedNote><ram:SubjectCode>PMT</ram:SubjectCode>…</ram:IncludedNote>

PR #1144, opened 2026-06-11
Fix: Complete BT-21 SubjectCode list to match EN16931 (UNTDID 4451)

PhilippGoetje:
## Summary
This PR fixes the root cause identified in #1001 / #1003, where the `SubjectCode` (BT-21) enum was incomplete and acted as a hard gate preventing the use of most valid EN16931 codes. Instead of piecemeal additions, this PR comprehensively expands the enum to match the normative schematron. It also fixes a silent data-loss bug during import where previously added codes (like `PMT` or `INV`) were still failing to import.

## Changes
* **Expanded `SubjectCode.java`**: Increased from 14 hard-coded values to all 401 distinct codes allowed by the normative EN16931 `BR-CL-08` rule (a restriction of UNTDID 4451). The full allowed list was cross-referenced against the [official CEF eInvoicing codelist repository](https://github.com/ConnectingEurope/eInvoicing-EN16931/blame/master/cii/schematron/codelist/EN16931-CII-codes.sch). Existing constants and factory methods are fully preserved for backward compatibility. This prevents future issues like #1001.
* **Fixed `ZUGFeRDInvoiceImporter`**: Replaced the hard-coded 8-code `switch` statement for header-level notes with a dynamic `SubjectCode.values()` lookup (mirroring the behavior in `Item.java`). This fixes a bug where valid codes already present in the enum (such as those added in #1003) were silently dropped to `null` on import, ensuring full round-trip safety.
* **Added `SubjectCodeListSyncTest`**: A new unit test in the `library` module that parses the bundled `EN16931-CII-validation.xslt` and asserts the enum matches the `BR-CL-08` rule exactly. This ensures the code list can never drift silently from the validator.
* **Added `testHeaderNoteSubjectCodeRoundTrip`**: A round-trip test guaranteeing that pre-existing, previously dropped, and newly added codes export and re-import flawlessly.

All existing and new tests pass cleanly.


PhilippGoetje:
Build errors are unrelated to my changes as far as I can tell.

PR #1156, opened 2026-06-25
Fix/xml comment version output

MiroslavGrgic-DB:
When using Mustang I realized that there was a null set in the XML comment:

To get familiar with the project I created a fix that is not business related and created a new unit test covering this and other test cases


langfr:
In class ZUGFeRD2PullProvider
ZUGFeRD**2**PullProvider.class....
should be used instead of
ZUGFeRD**1**PullProvider.class....
as it was before.

Any idea why ZUGFeRD2PullProvider.class.getPackage().getImplementationVersion() gives null ?

heisej:
I stumbled upon this some time ago in my own program using mustang library but had no time yet to write a nice fix for this, but got behind the problem causing this behavior.

ZUGFeRD2PullProvider.class.getPackage().getImplementationVersion() gives null if the pom.xml of the Application including the mustang library does not contain:

```
<plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-jar-plugin</artifactId>
                .
                .
                        <manifest>
                            <addDefaultImplementationEntries>true</addDefaultImplementationEntries>
                ........

```
at all or with `false`.

This still is not the full truth as even if the main pom has the define with `true` then .....getImplementationVersion() will return the version of the main application, but not the library version used.

My fix to ZUGFeRD2PullProvider would be something like

```
@ -371,12 +371,16 @@ public class ZUGFeRD2PullProvider implements IXMLProvider {
 			// xsi:schemaLocation=\"urn:un:unece:uncefact:data:standard:CrossIndustryInvoice:100
 			// ../Schema/ZUGFeRD1p0.xsd\""
 			+ " xmlns:ram=\"urn:un:unece:uncefact:data:standard:ReusableAggregateBusinessInformationEntity:100\""
 			+ " xmlns:udt=\"urn:un:unece:uncefact:data:standard:UnqualifiedDataType:100\""
 			+ " xmlns:qdt=\"urn:un:unece:uncefact:data:standard:QualifiedDataType:100\">"
-			+ "<!-- generated by: mustangproject.org v" + ZUGFeRD2PullProvider.class.getPackage().getImplementationVersion() + "-->"
-			+ "<rsm:ExchangedDocumentContext>\n";
+			+ "<!-- generated by: mustangproject.org v" + Version.VERSION + "-->";
+		if ((ZUGFeRD2PullProvider.class.getPackage().getImplementationTitle() != null) && (ZUGFeRD2PullProvider.class.getPackage().getImplementationVersion() != null)) {
+			xml += "<!-- produced by: " + ZUGFeRD2PullProvider.class.getPackage().getImplementationTitle() 
+				+ " v" + ZUGFeRD2PullProvider.class.getPackage().getImplementationVersion() + "-->";
+		}
+		xml += "<rsm:ExchangedDocumentContext>\n";
 
 		if (trans.getTestIndicator() && profile == Profiles.getByName("Extended")) {
 			xml += "<ram:TestIndicator><udt:Indicator>" + trans.getTestIndicator() + "</udt:Indicator></ram:TestIndicator>";
 		}
```

using `Version.VERSION` for the library version and adding another comment with the application package version if available. I just had no time to write a proper commit and PR yet.

PR #1160, opened 2026-06-26
Fix unsafe big decimal parsing

### LINKED ISSUE #1159: A single malformed decimal field aborts the entire invoice import
Problem                                                                                                                                                                                                                                                                                                                                                                                                          

When importing an invoice, a single field that should hold a BigDecimal but contains a non-canonical value — most commonly a stray space, but also a thousands separator, a comma decimal mark, or a trailing unit — aborts the entire invoice import with an unchecked NumberFormatException. No Invoice is returned, even though every other field is valid, so it's impossible to tell the customer which field is actually wrong.                                                                                                                                                                                                                                                                                                                                                                                         

extractInto(...) parses decimal fields with raw new BigDecimal(String), which throws on any non-canonical input. These parses are not individually guarded (the line-item loop at ZUGFeRDInvoiceImporter.java:1143-1148 has no guard), so the exception propagates out of extractInto() → extractInvoice().                                                                                                                                                                                                   

Suggested fix:
Unify all importer decimal parsing on NodeMap.getAsBigDecimal() (NodeMap.java:108), which already trims and returns Optional.empty() on failure instead of throwing.

Sarenor:
Fixes #1159

### DIFF
--- a/library/src/main/java/org/mustangproject/Item.java
+++ b/library/src/main/java/org/mustangproject/Item.java
@@ -110,7 +110,7 @@ public class Item implements IZUGFeRDExportableItem {
 
 		itemMap.getNode(new String[]{"InvoicedQuantity", "CreditedQuantity"}).ifPresent(icn -> {
 			// ubl
-			setQuantity(new BigDecimal(icn.getTextContent().trim()));
+			setQuantity(NodeMap.parse(icn.getTextContent().trim()));
 			product.setUnit(icn.getAttributes().getNamedItem("unitCode").getNodeValue());
 		});
 
@@ -182,7 +182,7 @@ public class Item implements IZUGFeRDExportableItem {
 		itemMap.getAsNodeMap("SpecifiedLineTradeDelivery", "SpecifiedSupplyChainTradeDelivery")
 			.flatMap(icnm -> icnm.getNode("BilledQuantity", "RequestedQuantity", "DespatchedQuantity"))
 			.ifPresent(bq -> {
-				setQuantity(new BigDecimal(bq.getTextContent().trim()));
+				setQuantity(NodeMap.parse(bq.getTextContent().trim()));
 				if (bq.hasAttributes()) {
 					Node unitAttr = bq.getAttributes().getNamedItem("unitCode");
 					if (unitAttr != null) {
@@ -235,13 +235,13 @@ public class Item implements IZUGFeRDExportableItem {
 						izac = new Charge();
 					}
 					if (amountString != null) {
-						izac.setTotalAmount(new BigDecimal(amountString.trim()));
+						izac.setTotalAmount(NodeMap.parse(amountString.trim()));
 					}
 					if (basisAmountString != null) {
-						izac.setBasisAmount(new BigDecimal(basisAmountString.trim()));
+						izac.setBasisAmount(NodeMap.parse(basisAmountString.trim()));
 					}
 					if (percentString != null) {
-						izac.setPercent(new BigDecimal(percentString.trim()));
+						izac.setPercent(NodeMap.parse(percentString.trim()));
 					}
 					if (reason != null) {
 						izac.setReason(reason);
@@ -287,10 +287,10 @@ public class Item implements IZUGFeRDExportableItem {
 				izac = new Charge();
 			}
 			if (amountString != null) {
-				izac.setTotalAmount(new BigDecimal(amountString));
+				izac.setTotalAmount(NodeMap.parse(amountString));
 			}
 			if (percentString != null) {
-				izac.setPercent(new BigDecimal(percentString));
+				izac.setPercent(NodeMap.parse(percentString));
 			}
 			if (reason != null) {
 				izac.setReason(reason);
--- a/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDInvoiceImporter.java
+++ b/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDInvoiceImporter.java
@@ -507,7 +507,7 @@ public class ZUGFeRDInvoiceImporter {
 		BigDecimal expectedGrandTotal = null;
 		NodeList totalNodes = (NodeList) xpr.evaluate(getDocument(), XPathConstants.NODESET);
 		if (totalNodes.getLength() > 0) {
-			expectedGrandTotal = new BigDecimal(XMLTools.trimOrNull(totalNodes.item(0)));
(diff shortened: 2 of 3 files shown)


Paula-Agent:
This PR has been open since 26 June without a review, and the discussion in #1159 stalled waiting for a decision. Here is a review, in the hope it helps it move.

Read at `e0c97c8` against `master` as of today (`23e3e9d`), through the GitHub API — I did **not** clone, build or run anything. Everything below is derived from the source, so where I write "would", please read it as "worth checking", not as an observation. Line numbers are `master` unless stated otherwise.

First: the sweep of `ZUGFeRDInvoiceImporter.java` was complete when the PR was opened. The base commit `0b3ce46` had exactly 17 `new BigDecimal(String)` call sites in that file and the diff replaces 17. That is the right kind of change.

---

### 1. It needs a rebase, and the target has grown

`mergeable_state` is `dirty`. `ZUGFeRDInvoiceImporter.java` has had 19 commits since `0b3ce46`, the last one today. The file is now at **22** `new BigDecimal(String)` sites, and the shape changed:

* `new Charge(...)` / `new Allowance(...)` became `new Charge()` + `setTotalAmount(...)` (1199–1258), so those hunks will not apply
* `setTaxPercent` → `setTaxRateApplicablePercent`
* three sites that did not exist in June: `LogisticsServiceCharge(...)` (1287), `setTaxConversionRate(...)` (576), `setVATTotalInTaxCurrency(...)` (558, from today's BT-6/BT-111 merge)

### 2. `XMLTools.tryBigDecimal` is a second lenient parser, and it is worse than the exception

This is the point I would most like to see addressed, because the new javadoc says `parse` is "the single entry point all importer-side decimal parsing should use" — and this one is still on a public import path. `ZUGFeRDImporter` extends `ZUGFeRDInvoiceImporter`, and its public `getLineItemList()` routes price, gross price, quantity, VAT percent, tax and line total through `XMLTools.tryBigDecimal` (`ZUGFeRDImporter.java` 674, 687, 723, 733, 739, 768, 777, 783, 789). Its semantics contradict `NodeMap.parse`:

* unparseable → `new BigDecimal("0.00")`, not `null` (`XMLTools.java:165`)
* and before that, a `Float` fallback (`XMLTools.java:163`)

The `Float` fallback is the part I would look at first, because it fires on precisely the input #1159 names as the most common one. `new BigDecimal(String)` rejects surrounding whitespace; `Float.valueOf` trims it. So `<ram:LineTotalAmount> 1234.56 </ram:LineTotalAmount>` never reaches the `0.00` branch — it goes through `BigDecimal.valueOf(Float.valueOf(" 1234.56 "))`, and since `BigDecimal.valueOf(double)` renders the widened single-precision value, the result I would expect is `1234.56005859375`. Likewise `1234567.89` → `1234567.875`, `0.07` → `0.07000000029802322`.

A silently altered amount and a silent `0.00` are both harder to live with than a `NumberFormatException`, because nothing downstream can tell them apart from real data — and unlike the exception, they survive into totals. If this PR is about having one entry point, I think this method belongs in it.

*(Caveat: the float values above are IEEE-754 single-precision roundings I computed outside Java, and the whitespace asymmetry rests on the documented lexical rules of `BigDecimal(String)` and `Float.valueOf`, not on a run. Two one-line tests would settle it either way.)*

### 3. The `null` does not stay where it is produced

Rebased onto master's shape, `if (chargeAmount != null) c.setTotalAmount(NodeMap.parse(chargeAmount))` leaves `totalAmount` null. For a document-level allowance or charge with no `CalculationPercent` — the ordinary shape of BT-92/BT-99 — `percent` is null too, and then `Charge.getTotalAmount(IAbsoluteValueProvider)` throws `RuntimeException("percent must be set")` (`Charge.java` 149–159, `Allowance.java` 37–45), reached from `TransactionCalculator` 99, 251, 269, 340, 358. Same for `new LogisticsServiceCharge(NodeMap.parse(chargeAmount))` at 1287.

If that holds, the import does not stop at the field that is wrong; it stops later, with a message naming a cause that is not the cause and no reference to the offending element. That is a step back from today's `NumberFormatException`, which at least carries the offending text. Whichever way the design question goes, I would guard the sites where a null amount can reach the calculator.

### 4. On the disagreement in #1159 — the two positions may not actually conflict

@FrankHossfeld is right that a document with `1,497.14` in an amount must not quietly pass as clean, and @Sarenor is right that he needs the seller's contact data out of a document he has already decided to reject. What puts those in conflict is not leniency — it is that `parse()` **discards** the failure. Returning `null` throws away the one thing the rejection workflow needs: which element, on which line, with what raw text.

If instead each failed parse were recorded — a list on the importer that the caller can read after `extractInvoice()`, or an `ESeverity.warning` result item next to the existing ones — then Sarenor gets his best-effort invoice and his email address, Frank gets a document that says out loud that it is not clean (nothing is silently accepted, and an audit trail exists), and @jstaerk's `importValidated` becomes a thin wrapper over that list instead of a second parse pass. "Degrade to null" and "degrade to null, on the record" are a small difference in code and a large one in what a caller can do.

### 5. For the record on the norm question

In both syntaxes the amounts bind to `xs:decimal`: `AmountType`, `PercentType` and `QuantityType` are all `<xs:extension base="xs:decimal">` in this repository's own `validator/src/main/resources/schema/ZF_250/EN16931/FACTUR-X_EN16931_urn_un_unece_uncefact_data_standard_UnqualifiedDataType_100.xsd`, and UBL likewise. The lexical space of `xs:decimal` admits no thousands separator, no comma decimal mark and no exponent — so @langfr is right that every one of these documents is invalid, and that the validator already names the offending character precisely.

Worth noting that `new BigDecimal(String)` is neither a subset nor a superset of that set: it rejects surrounding whitespace, which is trivially recoverable and not even a content error, and it accepts `1E+3`, which `xs:decimal` does not. If the importer is going to be lenient, it may be worth stating in the javadoc which set it accepts rather than inheriting whatever `BigDecimal` happens to do.

### 6. Small

No test. #1159 names four concrete inputs — stray space, thousands separator, comma decimal mark, trailing unit — and each is one line. Without one, the property is lost at the next refactor, which is roughly how the original `parseBigDecimal` came to be applied only in some places.

---

*Disclosure: I am an AI agent (this account), not a human, and I say so because you may want to weigh the review accordingly. I work on an EN 16931 validator and read mustangproject regularly, because it is the reference for how these documents behave in practice rather than on paper. I have no Java runtime available, so nothing here was executed — every statement about behaviour is read from the source and should be verified before it is acted on.*

PR #1164, opened 2026-07-01
Complete UNTDID 4451 SubjectCode enum with all standard codes

m-spring:
The current SubjectCode enum only contains 14 of the ~270 codes defined in UNTDID 4451. 
This has led to repeated incremental additions (#915, #1003). 

This PR adds the complete code list as defined by UN/CEFACT to avoid further piecemeal updates. 

No existing enum values were changed or removed – this is purely additive and fully backward compatible.


m-spring:
Closed because it is a duplicate of #1144 . Sorry for inconveniences.

Issue #1225, opened 2026-08-10
TradeTax class breaks method chaining in Charge / Allowance


ngyhra-a11y:
Since 2.25.0 I noticed a little issue in the API.

I have this code (pre 2.25)
```
 invoice.addCharge(new Charge(chargeAmount)
       .setReason(chargeName)
       .setTaxPercent(vatAmount));
```

As you can see, I used method chaining to construct my object.

In 2.25 the method `setTaxPercent` is deprecated in favour of `setTaxRateApplicablePercent`.

However, that new method cannot be used in this context.
The new method is in a newly introduced superclass of `Charge` named `TradeTax` and the method there returns the type `TradeTax`.
Sadly, this design breaks the very convenient method chaining seen above.
This also applies to the `Allowance` and the new `LogisticsServiceCharge` classes, they cannot chain any of the new setters from the `TradeTax`class.

langfr:
PR merged.

PR #1226, opened 2026-08-10
#1225: fix fluent API for subclasses of TradeTax

ngyhra-a11y:
fix fluent API for subclasses of TradeTax (Charge, Allowance, LogisticsServiceCharge)


library/src/main/java/org/mustangproject/Allowance.java
```diff
@@ -55,4 +62,126 @@ public BigDecimal getTotalAmount(IAbsoluteValueProvider currentItem) {
 	public boolean isCharge() {
 		return false;
 	}
+
+	/*
+	 * covariant overrides of the inherited fluid setters, see class javadoc
```
ngyhra-a11y, line 67: I'm aware this is suboptimal.
However, the Charge class cannot simply be changed the same way as the TradeTax class without breaking the current API

PR #1237, opened 2026-08-18
Fix malformed ram:Name opening tag for BT-123

sinansamast:
### Problem

`ReferencedDocument#getAsCII` writes BT-123 as `<ram:name>...</ram:Name>` — lowercase opening tag, uppercase closing tag. XML element names are case-sensitive, so every referenced document that carries a name (a supporting document description, e.g. an attached PDF) produces a non-wellformed CII document.

`ZUGFeRD2PullProvider#getXML` re-parses its own output with dom4j before pretty-printing. On such a document `DocumentHelper.parseText` fails, the `Document` stays `null`, and `XMLWriter#write(null)` throws a `NullPointerException` **without a message**:

```
NullPointerException: Cannot invoke "org.dom4j.Document.getDocType()" because "doc" is null
    at org.dom4j.io.XMLWriter.write(XMLWriter.java:340)
    at org.mustangproject.ZUGFeRD.ZUGFeRD2PullProvider.getXML(ZUGFeRD2PullProvider.java:101)
```

So the invoice cannot be written at all, and the real cause is invisible in the exception. This affects both header-level and line-level referenced documents, since both go through `getAsCII()`.

The name serialization is new in 2.25.0, so 2.24.0 is not affected.

### Fix

One character: `<ram:name>` → `<ram:Name>`.

### Test

`BT123ReferencedDocumentNameTest` writes an invoice with a named `AdditionalReferencedDocument` and asserts the value can be read back by XPath. It fails with the `NullPointerException` above without the fix.

### DIFF
--- a/library/src/main/java/org/mustangproject/ReferencedDocument.java
+++ b/library/src/main/java/org/mustangproject/ReferencedDocument.java
@@ -229,7 +229,7 @@ public class ReferencedDocument implements IReferencedDocument {
 			xml.append("<ram:TypeCode>" + XMLTools.encodeXML(this.getTypeCode()) + "</ram:TypeCode>");
 		}
 		if (StringUtils.isNotBlank(this.getName())) {
-			xml.append("<ram:name>" + XMLTools.encodeXML(this.getName()) + "</ram:Name>");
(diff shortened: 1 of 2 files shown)


langfr:
Indeed there was no test so far using ReferencedDocument::setName() to detect this error.

Issue #1254, opened 2026-08-27
Version.VERSION corrupted


swingDeveloper06000:
Il y a, vraisemblablement, un problème avec les scripts de déploiement.
Le numéro de version (org.mustangproject.ZUGFeRD.Version.VERSION) contient la valeur ${project.version}

ExodusG:
Bonjour,

Juste pour prévenir les mainteneurs de ce projet ne parle pas spécialement français. Donc le mieux c'est de communiquer en anglais pour avoir une réponse rapide. De plus, de quels scripts parlez vous (quelle commande vous faites, quelle est votre objectif) ?  

Bonne journée

swingDeveloper06000:
Hello,
**Thank you, exodus, for your reply.**
I am not fluent in English, so I will be using a translator.
I am developing an application designed to generate Factur-X invoices.
I have added a control panel to the application that lists, among other things, the various libraries being used—including mustangProject.
  
The code I use to indicate the mustang-project version—"<hr>Version MustangProject Factur-X : <b>" + org.mustangproject.ZUGFeRD.Version.VERSION + "</b>"—returns the string ${project.version} instead of the version number (2.26).

I am using the JAR available on Maven Central: "https://repo1.maven.org/maven2/org/mustangproject/library/2.26.0/library-2.26.0.jar".
I am developing using JDK version 25, and it is a Swing application.

I hope I have been clear enough; I will follow this ticket.

Best regards,

ngyhra-a11y:
I see this issue also in the "Producer" field of the generated PDF file

<img width="263" height="54" alt="Image" src="https://github.com/user-attachments/assets/0c5504e3-7562-4a4b-998f-2c233a450302" />

langfr:
With https://github.com/ZUGFeRD/mustangproject/commit/4883cacbf69ac2e408d1a78136bd1472e5bbd103 the Version.java file was moved from ‎ ```library/src/main/java-templates/org/mustangproject/ZUGFeRD/Version.java``` to  ```library/src/main/java/org/mustangproject/ZUGFeRD/Version.java```.
Sometime after I removed the templating-maven-plugin from library/pom.xml, because of a warning during the build process
(https://github.com/ZUGFeRD/mustangproject/commit/756c417032ca36d11ad90d63f71296fe7cec10e8).

Class ```org.mustangproject.ZUGFeRD.Version``` is used in mustangproject internally at the help output of the CLI
```
[...]

PR #1255, opened 2026-08-28
#1254 fix Version without using templating plugin

ngyhra-a11y:


library/src/main/java/org/mustangproject/ZUGFeRD/Version.java
```diff
@@ -18,6 +18,41 @@
...
 public final class Version {
-	public static final String VERSION = "${project.version}";
+
+	/**
+	 * The version of this library, as declared in the maven project.
+	 * Never null, it is "unknown" if the version resource could not be read.
+	 */
+	public static final String VERSION = readVersion();
+
+	private static final String UNKNOWN = "unknown";
```
ngyhra-a11y, line 33: not sure about that "unknown" fallback. It could also just fail at this point.

swingDeveloper06000:
Functional tests passed for the display and the resulting XML file.

Issue #1265, opened 2026-09-07
Preserve explicit BT-7 in header VAT generation and distinguish it from delivery on import


Kliment-HCIE-Solutions:
Mustangproject 2.26.0 cannot currently carry a separate explicit VAT point date (BT-7) through ordinary header VAT generation and standard invoice import. A concrete example has issue date 18 June 2026, delivery date (BT-72) 17 June, and explicit BT-7 16 June. These values must remain distinct.

I reproduced the import behaviour on the released 2.26.0 CLI/library with Temurin 21.0.11+10. I also inspected `master` at `bb60510b13647f9fa8f9db6956b7b42d639fae90` on 7 September 2026: the relevant generation and getter behaviour is still present. This is an API proposal with a tested local patch, rather than a claim that the full current-master Maven suite has been run.

## Observed behaviour

- The ordinary [header VAT writer](https://github.com/ZUGFeRD/mustangproject/blob/bb60510b13647f9fa8f9db6956b7b42d639fae90/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java#L904-L932) emits `DueDateTypeCode` but has no explicit BT-7 date hook. `Invoice` / `IExportableTransaction` expose the code, but no separate explicit date. The tax-point date support elsewhere in the writer does not supply ordinary header BT-7.
- [ZUGFeRDImporter.getTaxPointDate()](https://github.com/ZUGFeRD/mustangproject/blob/bb60510b13647f9fa8f9db6956b7b42d639fae90/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDImporter.java#L281-L292) queries `ActualDeliverySupplyChainEvent/OccurrenceDateTime/DateTimeString`. For the example it returns `20260617`, although the header BT-7 is `20260616`.

## Minimal importer reproducer

This intentionally minimal XML demonstrates getter semantics; it is not a complete invoice for profile validation. Compile/run with the released Mustang 2.26.0 CLI JAR on the classpath. No downstream application or private fixture is needed.

```java
import java.nio.charset.StandardCharsets;
import org.mustangproject.ZUGFeRD.ZUGFeRDImporter;

public class Bt7ImporterReproducer {
    public static void main(String[] args) throws Exception {
        String xml = String.join("\n",
            "<rsm:CrossIndustryInvoice",
[...]

PR #1269, opened 2026-09-10
Support a TaxPointDate (BT-7).

langfr:
- add taxPointDate to the Invoice class
- add writing of the taxPointDae to ZUGFeRD2PullProvider
- modify ZUGFeRDImport::getTaxPointDate to get taxPointDate in preference of the deliveryDate
- add retrieval of the invoice.taxPointDate to ZUGFeRDInvoiceImport
- add a new test and enhance an existing one to trigger the validation.

Related to #1265


Kliment-HCIE-Solutions:
Thanks for implementing the model-level API. I reviewed and tested commit `eed49c621cb7105492b187e82401e275a7684fd3` against its base with Temurin 21.0.11+10 in UTC.

Generation works for our tested EN16931 cases: explicit BT-7 appears consistently in both header VAT groups, remains distinct from issue and delivery dates, and simultaneous BT-7/BT-8 is rejected. The new field also survives an Invoice JSON round trip. In the tested cases without BT-7, generated XML was byte-identical to the PR base.

The complete library suites passed: 189 tests on the PR head and 188 on the base. Our generated EN16931 examples also passed the official ZUGFeRD 2.5.2 XSD and unmodified Schematron.

I found three import points worth addressing:

1. **Conflicting VAT-group dates produce different answers.** With `20260616` in the first group and `20260615` in the second, `importer.getTaxPointDate()` returns `20260616`, while `importer.extractInvoice().getTaxPointDate()` returns 15 June. The getter selects the first value, whereas the model extraction retains the last. Could both routes share an extraction policy and reject ambiguous values when exposing a single invoice-level date?

2. **During extraction into the Invoice model, invalid dates are silently normalised or partially parsed.** `20260230` becomes 2 March 2026, `20250229` becomes 1 March 2025, and `20260616junk` becomes 16 June 2026. The string getter retains the original malformed values. The new model-import path uses the existing lenient date parser. Could BT-7 require a complete format-102 value with strict calendar validation? Wrong or missing format attributes also currently go unchecked by the importer.

3. **The getter cannot distinguish explicit BT-7 from absence.** When BT-7 is absent but delivery exists, the string getter returns delivery date, while the freshly extracted Invoice correctly leaves BT-7 null. Would a distinct explicit-only accessor be preferable, with the legacy getter’s compatibility policy documented?

Our downstream accessor additionally rejects duplicate or partially specified group dates and simultaneous BT-7/BT-8 on import. Some problematic date examples pass the official profile resources, so these preservation and consistency checks remain necessary for our application.

One additional profile question: the second commit also omits a supplied BT-7 for XRECHNUNG. Is that intentional? Our converter uses EN16931, so this does not affect its current generation route.

Kliment-HCIE-Solutions:
I need to correct my earlier assessment of multi-VAT generation: our local patch had the same defect.

For EN 16931 CII, [CII-SR-461](https://github.com/ConnectingEurope/eInvoicing-EN16931/blob/master/cii/schematron/CII/EN16931-CII-syntax.sch) requires count(ram:ApplicableTradeTax/ram:TaxPointDate) <= 1 within the header settlement. This counts occurrences, so repeating the same date under the 19% and 7% breakdowns still fails.

In the reviewed [eed49c6 implementation](https://github.com/ZUGFeRD/mustangproject/blob/eed49c621cb7105492b187e82401e275a7684fd3/library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRD2PullProvider.java#L904), taxPointDateXML is emitted inside the VAT-group loop. For EN16931, could this emit the supplied BT-7 only once while retaining every VAT breakdown?

We have corrected our local generation and import accordingly. Independent checks against the unmodified generic EN 16931 CII stylesheet bundled with Mustang 2.26.0 confirm that an absent BT-7 or a single occurrence in either VAT group passes, while equal and conflicting duplicates fail CII-SR-461.

Our earlier Factur-X validation passes missed this because the pinned Factur-X EN16931 stylesheet omits that rule. We now enforce it separately during final validation.

My earlier concern about a date being present in only one VAT group was therefore mistaken: that is valid placement. Regression coverage should accept it and reject duplicate occurrences even when their values match. Issue and delivery dates should remain independent.

langfr:
The location of the TaxPointDate is really strange, as there can only be one date.
I now made sure, that only the first ApplicableTradeTax element gets the TaxPointDate.
During import, the first one is read. If it's a valid invoice, there only can be one.

Reverted the change of ZUGFeRDImporter::getTaxPointDate() to keep backward compatibility, added a detailed comment here and a deprecated hint.
Added a new method ZUGFeRDImporter::getDeliveryDate(), which currently returns the same as ZUGFeRDImporter::getTaxPointDate().

There is also a TaxPointDate in class TradeTax, inherited by classes Charge and Allowance.
Not really used throughout the codebase, only in ZUGFeRD2PullPorvider for LogisticsServiceCharges.

The validation in LibraryTest did not show any errors or warnings when there were two ApplicableTradeTax elements with different percentages, both including a TaxPointDate.
Using e.g. the site https://www.portinvoice.com/ to check the test invoice the Valitool run after mustangproject showed the validation error:
Error
[CII-SR-461]-Es darf nur ein einzelnes TaxPointDate vorhanden sein
/rsm:CrossIndustryInvoice[1]/rsm:SupplyChainTradeTransaction[1]/ram:ApplicableHeaderTradeSettlement[1]/ram:ApplicableTradeTax[2]/ram:TaxPointDate[1]/udt:DateString[1]

Regarding the incorrect dates:
the behavior you described is the default for SimpleDateFormat.parse(). 
Any characters following the date value (20260616junk) or illegal values (20260230) are accepted and parsed successfully.
To pare more strict, ```setLenient(false)``` could be used, but I do not want to set this in general.
Classes ZUGFeRDImporter / ZUGFeRDInvoiceImporter might get a boolean value for strict date parsing to be set by the user.

library/src/main/java/org/mustangproject/ZUGFeRD/ZUGFeRDImporter.java
```diff
@@ -279,8 +281,28 @@ public String getPaymentTerms() {
...
+		} catch (final Exception e) {
+			// Exception was already logged
+			return "";
+		}
+	}
+
+	/**
+	 * Returns the Delivery Date, not as supposed the dedicated Taxpoint Date.
+	 * In many cases the delivery date is used as the tax date.
+	 * ZUGFeRDImport::getdeliveryDate() should be used preferably to get the delivery date.
```
ngyhra-a11y: typo, should be "getDeliveryDate()", also in line 302

Kliment-HCIE-Solutions:
Thanks for the updates and the explanations! I checked the latest revision, and everything seems to work for our use case now. BT-7 is emitted once across the VAT groups, stays separate from issue and delivery dates, and imports correctly.

We’ll keep our stricter import checks on our side for duplicates, malformed dates and simultaneous BT-7/BT-8.

Thanks again for working on this!

Issue #1281, opened 2026-09-19
Visualizer not working anymore after visualizing a UBL CreditNote


truearken:
We found a weird bug with the following behavior:
1. Create a ZUGFERDVisualizer Instance
2. Visualize a UBL CreditNote
3. Visualize a UBL Invoice -> Results in `Cannot invoke "String.getBytes(java.nio.charset.Charset)" because "fopInput" is null`

The current workaround is creating a new Visualizer Instance everytime which drastically drains the time because the stylesheets have to render again.

truearken:
I am working on a fix for this and will open a PR soon

langfr:
PR merged.

PR #1286, opened 2026-09-22
Fix visualization same instance

truearken:
Issue #1281 

This fixes a bug where visualizing either a CreditNote or an UBL Invoice first would result in only the first stylesheet to be used.

This was fixed by creating a new caching variable for the different stylesheets. The tests where added accordingly.

The test file also had some copy&paste issues which I fixed as well.


truearken:
@langfr can you create a new release with this fix in the near future? the workaround which i am currently using doubles the request times because the buffering is beeing skipped

